PR #2, opened 2020-02-09
ui-components package for "core components"

nathanstilwell:
Creating "core components" package to publish to [@cockroachlabs/ui-components](https://www.npmjs.com/package/@cockroachlabs/ui-components).

In this PR I would like to accomplish two things:
• Setup a components library project with tools for building, testing, and publishing
• Create an example component as a reference (but not suitable for production)

This pull request seeks to establish the following norms:

1. Our _component library_ will consist of React components using CSS Modules written in Sass. We will continue to evaluate [astroturf](https://github.com/4Catalyzer/astroturf) as an option for generating components and CSS modules, but will not make its usage strictly necessary.

2. We will leverage [Webpack](https://webpack.js.org/) and [Babel](https://babeljs.io/) to bundle JS and CSS Modules, but it is important to export modules in such a way that they can be _tree-shaken_ by the consuming application.

3. All components in this library should be tested. We have chosen [Jest](https://jestjs.io/) and [Enzyme](https://airbnb.io/enzyme/) for this purpose.

4. All components should have their API documented in a local README.


.gitignore
```diff
@@ -0,0 +1,36 @@
...
+*.tgz
+
+# Yarn Integrity file
+.yarn-integrity
+
+# dotenv environment variables file
+.env
+.env.test
+
+# parcel-bundler cache (https://parceljs.org/)
```
laurenbarker: Just curious, where did this `.gitignore` come from?

packages/ui-components/src/Sample/Sample.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { FunctionComponent } from "react";
+import classNames from "classnames/bind";
```
laurenbarker: It might be worth noting on the main README that we need to use the [bind version of classnames](https://github.com/JedWatson/classnames#alternate-bind-version-for-css-modules) with css modules.

packages/ui-components/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "type": "git",
+    "url": "https://github.com/cockroachdb/ui"
+  },
+  "main": "dist/main.js",
+  "scripts": {
+    "build": "webpack --display-error-details",
+    "clean": "rm -rf  ./dist/*",
+    "test": "jest --watch",
+    "watch": "webpack",
+    "test:once": "jest"
```
laurenbarker: Does running jest trigger type checking?

packages/ui-components/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "classnames": "^2.2.6",
+    "enzyme": "^3.11.0",
+    "enzyme-adapter-react-16": "^1.15.2",
+    "identity-obj-proxy": "^3.0.0",
+    "jest": "^25.1.0",
+    "jest-environment-enzyme": "^7.1.2",
+    "jest-enzyme": "^7.1.2",
+    "lodash": "^4.17.15",
+    "node-sass": "^4.13.1",
+    "react": "^16.12.0",
```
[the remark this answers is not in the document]
nathanstilwell: Nice catch. You are totally right.

packages/ui-components/src/Sample/README.md
```diff
@@ -0,0 +1,32 @@
+[back to components](../README.md)
```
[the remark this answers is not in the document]
nathanstilwell: Storybook (if I remember correctly) has a plugin that allows you to import READMEs to be viewed alongside the rest of the _story_. I think Storybook will be an invaluable enhancement to the documentation, but I would like to think at a minimum an engineer could get some docs along side the code. Storybook (and even Typescript annotations) may eventually make it unnecessary, but it's just kind of an old habit for me.

If you find this overkill or too cumbersome, I'm open to other ideas.

packages/ui-components/src/Sample/Sample.tsx
```diff
@@ -0,0 +1,29 @@
...
+  tint?: "default" | "red" | "green" | "blue";
+}
+
+const cx = classNames.bind(styles);
+
+const Sample: FunctionComponent<SampleProps> = props => {
+  const { content, tint, children } = props;
+  const classnames = cx("sample", tint);
+
+  if (!isUndefined(children)) {
```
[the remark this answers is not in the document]
nathanstilwell: Old habits die hard. I'll wait to have the _lodash conversation_ until I'm using it for something more substantial. It does feel silly to have an entire library for `isUndefined`.

packages/ui-components/webpack.config.js
```diff
@@ -0,0 +1,68 @@
...
+        loader: "source-map-loader",
+      },
+    ],
+  },
+
+  plugins: [
+    new WebpackBar({
+      name: "ui-components",
+      color: "cyan",
+      profile: true,
```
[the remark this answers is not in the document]
nathanstilwell: webpack-bar

There's not a lot going on yet, but here's the idea. 😄

.gitignore
nathanstilwell: Generated by Github I think.

packages/ui-components/package.json
nathanstilwell: No, I don't think so. I'll try to work that in somewhere. When do you think Type checking should happen? On build or on test? both?

packages/ui-components/src/Sample/README.md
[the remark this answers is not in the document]
laurenbarker: Oh cool! I didn't know there was a [readme plugin](https://github.com/tuchk4/storybook-readme). 

I was imagining some docs in READMEs and some in storybook but as long as we can see all of the README docs on the storybook pages (once they exist) then we just have to worry about consistency. We can figure out what pattern we want later 👍

packages/ui-components/src/Sample/Sample.tsx
nathanstilwell: Would that be to provide context for Authorship (for contributors)? I had considered introducing a CONTRIBUTING.md in this PR, but thought I might add it when actually introducing a real component. Do you think it's worth establishing those norms now?

Other things I would consider needing to go into CONTRIBUTING would be things like test setup, mentioning the option for Astroturf, what constitutes minimal documentation. What do you think?

packages/ui-components/package.json
nathanstilwell: You know what else I forgot?! **Linting**! I definitely need to get our ESLint config choices integrated in here.

laurenbarker: Ah right, linting :). Feel free to do a follow-up PR if that's easier. It seems like most people (the internet people) like when the build doesn't fail if type checking fails. As long as it fails in CI, I don't have a preference though.

packages/ui-components/src/Sample/Sample.tsx
laurenbarker: Adding to `CONTRIBUTING.md` would also work or even a PR template. We can wait and see if it becomes a problem though. It might not even be an issue.

.gitignore
laurenbarker: Nope! Was just wondering if this was the GitHub generated one. The norms change so quickly haha

packages/ui-components/package.json
nathanstilwell: So after going down the rabbit hole a little bit, I found out that indeed Jest **is** doing type checking since we are using the `ts-jest` [transform](https://github.com/cockroachdb/ui/blob/52e2e022d69c39b06f53f805f661f8cc4ab3c888/packages/ui-components/jest.testing.config.js#L12). I am hooking up linting and [type-checking](https://github.com/azz/jest-runner-tsc) anyway, since that seems to check the [definitions file](https://github.com/cockroachdb/ui/blob/52e2e022d69c39b06f53f805f661f8cc4ab3c888/packages/ui-components/src/declaration.d.ts) for issues.

Getting Jest, Webpack, Typescript, Babel, and whoever else to work together seemed hard enough that I don't really want to change anything. So let me know if at some point in the future the type checking gets too cumbersome.

packages/ui-components/src/declaration.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "*.scss" {
+  const content: { [className: string]: string };
```
j-low: I'm just getting initiated to CSS modules as I review this...for the sake of my understanding: this is what enables the access-CSS-class-as-object-property functionality of CSS modules, correct?

nathanstilwell: I think so (not exactly sure what you mean). This is what keeps Typescript from complaining about 

```import styles from "path/to/some.module.scss"```

Without it Typescript doesn't know what to do with this and throws an error like `Cannot find module 'some.module.scss'. ts(2307)` It seemed to flake out a little on my yesterday (probably because I was _futzing_ around with type checking), so if you see that error this, this is where the problem will be resolved.

j-low: Sorry that was poorly worded...I meant that it enabled the module behavior in the sense that it Typescript recognized it as such.

PR #3, opened 2020-02-12
Badge Component

nathanstilwell:
# Badge Component

Screen Shot 2020-02-15 at 10 04 32
Screen Shot 2020-02-15 at 10 11 01
Screen Shot 2020-02-15 at 10 05 02
Screen Shot 2020-02-15 at 10 05 18
Screen Shot 2020-02-15 at 10 06 43

Migrating the `Badge` component from Cockroach Cloud and adjusting the properties slightly. This component will reduce the API in the component current in [Admin UI](https://github.com/cockroachdb/cockroach/blob/master/pkg/ui/src/components/badge/badge.tsx), but I am applying the strategy of not including any functionality that is currently not needed in Cockroach implementations. 

I am posting samples above in lieu of a running example while our component catalog is still under development.

Since this is the first valid component in the library, this PR will remove the sample component.


packages/ui-components/src/Badge/Badge.module.scss
```diff
@@ -0,0 +1,45 @@
...
+  font-size: 12px;
+  line-height: 14px;
+  padding: crl-gutters(0.25) crl-gutters(0.75);
+  vertical-align: crl-gutters(0.5);
+  text-align: center;
+  white-space: nowrap;
+  width: fit-content;
+}
+
+.intent--neutral {
```
koorosh: @nathanstilwell , what do you think about following BEM naming conventions?

packages/ui-components/src/Badge/Badge.tsx
```diff
@@ -0,0 +1,52 @@
...
+  intent?: BadgeIntent;
+  transformCase?: BadgeCase;
+}
+
+const cx = classNames.bind(styles);
+
+const Badge: FunctionComponent<BadgeProps> = ({
+  content,
+  intent = "neutral",
+  transformCase = "uppercase",
```
koorosh: @nathanstilwell , it is possible to assign default properties with `defaultProps` static method
```javascript
const Badge: FunctionComponent<BadgeProps> = () => { ... }

Badge.defaultProps: Partial<BadgeProps> = {
  intent: "neutral",
  transformCase: "uppercase",
}
```

packages/ui-components/src/Badge/Badge.tsx
```diff
@@ -0,0 +1,52 @@
...
+  children,
+  ...props
+}) => {
+  const classnames = cx("badge", {
+    "intent--success": intent === "success",
+    "intent--danger": intent === "danger",
+    "intent--warning": intent === "warning",
+    "intent--neutral": intent === "neutral",
+    "transformCase--uppercase": transformCase === "uppercase",
+    "transformCase--none": transformCase === "none",
```
koorosh: @nathanstilwell , as a shortcut for assigning classes depending on props might be useful in case new `intent` or `transformCase` options will be added:
```javascript
const classnames = cx("badge", `intent--${intent}`, `transformCase--${transformCase}`)
```
Having default values, it is safe to assume that only valid values will be provided

packages/ui-components/src/Badge/Badge.tsx
laurenbarker: Great question, @koorosh! We decided against using `defaultProps` since it is will be [deprecated for function components](https://github.com/reactjs/rfcs/pull/107).

packages/ui-components/src/Badge/Badge.tsx
```diff
@@ -0,0 +1,52 @@
+import React, { FunctionComponent } from "react";
+import classNames from "classnames/bind";
+
+import styles from "./Badge.module.scss";
+
+type BadgeCase = "none" | "uppercase";
+type BadgeIntent = "neutral" | "success" | "warning" | "danger";
+
+export interface BadgeProps {
+  content?: string;
```
laurenbarker: Is `content` an option for backwards compatibility in Admin UI? Having two ways to define text content opens an opportunity for incorrect usage/unexpected behavior:

```js
<Badge content="This won't be used">This will be used</Badge>
```

nathanstilwell: Good point.

packages/ui-components/src/Badge/Badge.module.scss
nathanstilwell: I am choosing not to use BEM here as I'm trying to represent a different paradigm (classnames based on props rather than classnames based on DOM structure and state). I am going to change the format slightly (see [comment](https://github.com/cockroachdb/ui/pull/4#issuecomment-589183018)) to better align with [Astroturf conventions](https://github.com/4Catalyzer/astroturf#component-api) (changing from `${propName}--${propValue}` to `${propName}-${propValue}`).

packages/ui-components/src/Badge/Badge.tsx
nathanstilwell: I think this is a valid solution to tighten up this code. I'm proposing a utility to solve this [here](https://github.com/cockroachdb/ui/pull/4) and plan to refactor using that. I just left the _exploded_ version here to be explicit and perhaps illustrative.

PR #4, opened 2020-02-13
Object to classnames utility

nathanstilwell:
In order to create classes in CSS Modules that are less likely to
conflict, I'm wondering if a convention that might work would be to use
`[propName]-[propValue]` for css classes that will equate to presentational
changes based on props.

To accomodate this more clearly, I'm creating a utility to generate
these classnames from props (a util called `objectToClassnames`)


nathanstilwell:
@laurenbarker I disagree about the amount of complexity being introduced, but agree that the value being provided is purely sugar (it just saves time in typing). I will agree that at first blush, the value may not be immediately apparent by looking at the function so I have tried to express that value in the README. If there is anything I can do to make it more digestible, please let me know. At the end of the day, this is just a formatting utility and it's use is optional so I leave it to component authors to adopt or not.

That said, I did reflect on our discussion of keeping consistency with Astroturf and so I will update the default to be `${propName}-${propValue}` with a single dash instead of a double. This way the styles written in a `Component.module.scss` file and in an Astroturf `css` block are interchangeable.

PR #5, opened 2020-02-14
Add shared eslint config as a package

laurenbarker:
Since we now use the same eslint config in CockroachCloud, ui-components,
and other new packages in our ui repo it makes sense to pull the eslint
and prettier config into a shared location. This will ensure that all
of our react components are formatted in the same way.

To use in other packages/apps:
```
// eslintrc.json
{
  "extends": "@cockroachlabs/eslint-config"
}
```

Tested via https://github.com/cockroachlabs/managed-service/pull/2150


nathanstilwell:
One minor note, you may want to include an `.npmignore` file for the lock file. It doesn't really matter as lock files get ignored for transient dependencies.

PR #8, opened 2020-02-20
Icon Package

nathanstilwell:
Breaking the icon svgs out into their own package (@cockroachlabs/icons)
to be consumed either as components that render raw svgs or by the Icon
component in ui-components.

This package will only contain a folder of svg file an index.ts that exports them. Webpack using svgr will do all the work by bundling the svgs as React components.


packages/icons/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@cockroachlabs/icons",
+  "version": "0.1.0-prerelease.0",
+  "description": "Icons for Cockroach UI exported as React Components",
+  "files": [
+    "dist/main.js",
+    "dist/main.d.ts"
+  ],
+  "scripts": {
+    "build": "webpack --mode development",
```
laurenbarker: Does the mode matter?

packages/icons/webpack.config.js
```diff
@@ -0,0 +1,31 @@
...
+    path: path.resolve(__dirname, "dist"),
+    filename: "main.js",
+    library: "uiComponents",
+    libraryTarget: "umd",
+  },
+  module: {
+    rules: [
+      {
+        test: /\.svg$/,
+        use: ["@svgr/webpack"],
```
laurenbarker: Nice find!

packages/icons/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "build:watch": "webpack --watch --mode development",
+    "clean": "rimraf dist",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "author": "Nathan Stilwell <nathanstilwell@cockroachlabs.com",
```
laurenbarker: nit: don't think we need an author on our packages

packages/icons/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "license": "MIT",
+  "devDependencies": {
+    "@svgr/webpack": "^5.1.0",
+    "@types/react": "^16.9.20",
+    "copy-webpack-plugin": "^5.1.1",
+    "npm-run-all": "^4.1.5",
```
laurenbarker: Is this being used?

packages/icons/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "rimraf": "^3.0.2",
+    "webpack": "^4.41.6",
+    "webpack-cli": "^3.3.11"
+  },
+  "dependencies": {
+    "react": "^16.12.0"
```
laurenbarker: I think `react` should be in `devDependencies` in addition to `peerDependencies`. I'm not positive but I would imagine if it's in both `dependencies` and `peerDependencies` then it would get treated as a dependency.

packages/icons/package.json
nathanstilwell: nope

packages/icons/package.json
nathanstilwell: This does matter, in that I wasn't setting a mode in the webpack config. To simplify matters, I'm going to set the mode to `none` in the config and take out the switch.

PR #11, opened 2020-02-20
Publish ui-components v0.1.0

nathanstilwell:
- @cockroachlabs/ui-components@0.1.0

### DIFF
--- a/packages/ui-components/package-lock.json
+++ b/packages/ui-components/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@cockroachlabs/ui-components",
-  "version": "0.1.0-prerelease.1",
+  "version": "0.1.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


nathanstilwell:
This is a formality as I'm trying to work out the "release process" around publishing. Since we are publishing _manually_ (from a laptop using Lerna) PRs like these will simple serve as a record instead of pushing straight to master. In this case the package has already been [published to npm](https://www.npmjs.com/package/@cockroachlabs/ui-components/v/0.1.0)

laurenbarker:
If we want to go through PRs prior to publishing to npm we could:
* run [lerna version](https://github.com/lerna/lerna/tree/master/commands/version#readme) on branch which bumps the version but doesn't publish to npm
* submit a PR
* merge PR 
* run [lerna publish from-package](https://github.com/lerna/lerna/tree/master/commands/publish#bump-from-package) on master which publishes packages to npm whose versions don't match

Separating version and publish ensures that nothing gets published to npm that doesn't match master.

PR #12, opened 2020-02-20
let's try a github workflow

nathanstilwell:
This is a total experiment


nathanstilwell:
Ok, so this is a placeholder until we can get TeamCity setup and integrated with Github.

PR #14, opened 2020-02-21
updating CONTRIBUTING with feedback

nathanstilwell:


nathanstilwell:
Oh right. I have only recently switched to _fork and pull_ request and probably haven't published since then. Are you sure the switch option is `--git-remote [remote]`. I'm on Lerna version `3.20.2` and that option isn't documented. I haven't done a test publish yet, but I will and document it before I merge this PR.

laurenbarker:
The `--git-remote` flag does work on the latest version. That's what I used to publish the `eslint-config` package. It's only documented on the [version command](https://github.com/lerna/lerna/tree/master/commands/version#--git-remote-name) though.

PR #15, opened 2020-02-21
Add publishConfig so publishing scoped packages doesn't fail

laurenbarker:
In order to publish scoped packages, access must be set to public.

See https://github.com/lerna/lerna/tree/master/commands/publish#per-package-configuration for more details.


laurenbarker:
ran `lerna publish from-package` after merging to actually publish the package.

PR #19, opened 2020-04-07
Resolution for minimist vulnerability

nathanstilwell:
Upgrading `eslint` to resolve vulnerability in `minimist@0.2.0` (see
https://github.com/advisories/GHSA-vh95-rmgr-6w4m)

```
@cockroachlabs/eslint-config@0.1.2 /Users/nathanstilwell/Code/ui/packages/eslint-config
└─┬ eslint@6.8.0
  └─┬ mkdirp@0.5.5
    └── minimist@1.2.5
```

### DIFF
--- a/packages/eslint-config/yarn.lock
+++ b/packages/eslint-config/yarn.lock
@@ -9,14 +9,24 @@
   dependencies:
     "@babel/highlight" "^7.8.3"
 
+"@babel/helper-validator-identifier@^7.9.0":
+  version "7.9.5"
+  resolved "https://registry.yarnpkg.com/@babel/helper-validator-identifier/-/helper-validator-identifier-7.9.5.tgz#90977a8e6fbf6b431a7dc31752eee233bf052d80"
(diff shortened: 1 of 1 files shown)


packages/eslint-config/package.json
```diff
@@ -33,5 +33,8 @@
   },
   "publishConfig": {
     "access": "public"
+  },
+  "resolutions": {
```
laurenbarker, line 37: I'd prefer to bump a version instead of add a resolution if possible. Not sure which one `minimist` is a dependency of though...would have to check.

laurenbarker: Oh it's eslint

nathanstilwell: It's buried pretty deep and there's not a top-level dependency to upgrade.

laurenbarker: `yarn upgrade eslint` will update `minimist` (and some other stuff) in the lock file. Looks like [mkdirp fixed it](https://github.com/eslint/eslint/issues/13050#issuecomment-600311496)

nathanstilwell: yeah, I was about to take it back. I think we can just bump the version and fix it.

kenliu:
thanks again for taking on these dependency updates!

PR #41, opened 2020-05-03
ui-components: Icon Component

nathanstilwell:
# ui-components // Icon

Screen Shot 2020-05-13 at 16 03 03
Screen Shot 2020-05-13 at 16 02 52

Icon component is an React component interface for `@cockroachlabs/icons`.

#### Checklist


packages/icons/README.md
```diff
@@ -36,3 +36,54 @@ Each Icon component will render an `<svg />` element with an unbounded size and
   
   <CheckCircleFilled style={{ fill: "green", height: "24px", width: "24px" }} />
 ```
+
+## Icons
+
+Included icons in this package are 
+`ArrowLeft`,
```
laurenbarker, line 43: Is worried about having to maintain this list in several places. Having a single source of truth would be ideal. Could we export a list of icon names from the icon package?

packages/storybook-ui-components/stories/2-Icon.stories.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+export default {
+  title: "Icon",
+  component: Icon,
+  decorators: [withKnobs],
+};
+
+const IconDisplaySection = ({ children }) => (
+  <section
+    style={{
```
laurenbarker, line 14: Would be great if we were able to share these styles across stories so we could have consistent layouts. Might be worth a ticket if it doesn't already exist.

packages/storybook-ui-components/stories/2-Icon.stories.tsx
```diff
@@ -0,0 +1,150 @@
...
+  "Plus",
+  "Search",
+  "Spinner",
+  "Stack",
+  "Terminal",
+  "Time",
+  "User",
+  "World",
+];
+const sizes = ["large", "default", "small", "tiny"];
```
laurenbarker, line 92: Medium is missing

packages/ui-components/src/Icon/Icon.module.scss
```diff
@@ -0,0 +1,61 @@
...
+
+.tint-white {
+  fill: white;
+}
+
+.tint-neutral {
+  fill: #242a35;
+}
+
+.tint-inherit {
```
laurenbarker, line 59: Is `inherit` an option for tint?

packages/ui-components/src/Icon/Icon.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { FunctionComponent } from "react";
+import classNames from "classnames/bind";
+import get from "lodash/get";
+
+import * as Icons from "@cockroachlabs/icons";
+import styles from "./Icon.module.scss";
+
+import objectToClassnames from "../utils/objectToClassnames";
+
+type IconSize = "large" | "medium" | "default" | "small" | "tiny";
```
laurenbarker, line 10: I think we usually use t-shirt sizes. It's probably worth checking in with design (if you haven't already) to make sure they use the same names. I don't see different sizes laid out in the DS mocks.

packages/ui-components/src/Icon/Icon.tsx
```diff
@@ -0,0 +1,44 @@
...
+type IconTint =
```
laurenbarker, line 11: Where did these tints come from? Are they all in use? (purple seems like it shouldn't be in use)

packages/storybook-ui-components/stories/2-Icon.stories.tsx
laurenbarker: It might be easier to keep these in sync if they are exported from the component.

packages/ui-components/src/Icon/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Properties
+
+These are the properties that can be provided to the `Icon` component.
+
+### `iconName: string`
+
+The `iconName` prop should be a valid icon name from the [icons package](https://github.com/cockroachdb/ui/tree/master/packages/icons#icons). If an invalid iconName is supplied, the component will return `null`.
+
+### `size?: Enum("large", "medium", "small", "tiny")`
```
laurenbarker, line 27: I'd be more comfortable if prop options were documented by generated stories rather than us updating them every time we change the component.

packages/ui-components/src/Icon/README.md
```diff
@@ -0,0 +1,50 @@
...
+`small` | `12 x 12 `
+`tiny` | `8 x 8`
+
+### `tint?: Enum("blue", "green", "purple", "red", "white", "neutral", "inherit")`
+
+The `tint` prop will alter the fill color of the rendered icon. The color values of the possible tint property values are,
+
+tint     | fill color
+-------- | ----------
+`blue`   | `#0788ff`
```
laurenbarker, line 44: I don't think we should have hex values in the READMEs since they are (or rather will be soon) dependent on tokens.

packages/icons/README.md
nathanstilwell: Now that we have Storybook in a public place I think I will just make that the official list so we don't have to maintain a list in a README

packages/ui-components/src/Icon/Icon.module.scss
nathanstilwell: I think I meant it to be.

packages/ui-components/src/Icon/README.md
nathanstilwell: Are you suggesting I delete this README? Or that we use stories as the only component documentation?

laurenbarker: I think everything below "Properties" should be removed since it's not sustainable to maintain. Just having the properties documented in generated stories and by the code should cover all of our use cases. I don't have strong feelings on where the description and example should live but I imagine we would want to have a those in a single location as well. What do you think?

packages/storybook-ui-components/stories/2-Icon.stories.tsx
nathanstilwell: Anne and Rachel are working on _standardizing_ what our stories should look like. I agree with you about these ad hoc inline styles. I'm hoping their work will resolve the inconsistencies in the stories as well as the README conversations we've been having all over the place.

All that to say, I'm punting for now.

packages/storybook-ui-components/stories/2-Icon.stories.tsx
nathanstilwell: So I looked into this. I have exported the `IconSize` type, but since it is a [string literal type](https://www.typescriptlang.org/docs/handbook/advanced-types.html#string-literal-types) it can not be iterated over like an array. I could export an array of strings of the literal type, but it wouldn't really serve any other purpose except in creating this story. I also considered an [Enum](https://www.typescriptlang.org/docs/handbook/enums.html) which could be iterated over, but would then require an implementer to import the Enum in order to pass props to the component an I don't think is as ergonomic as using a string.

So my compromise is to export the custom type and then type this array as `Array<IconSize>`. while this doesn't solve the problem of missing types, it does error if you try to add something to the array that's not of type IconSize. Since these stories aren't likely to change very often I don't think the maintenance cost is very high here.

packages/ui-components/src/Icon/Icon.tsx
nathanstilwell: I took the props for this component from the existing Icon component in managed service. I will confirm with designer what sizes of icon are preferable.

As far as t-shirt sizes go, do you mean we usually use abbreviations? (like `l`, `m`, `s`, `xs`, etc) I find those to be confusing and wanted to avoid bike-shedding whether to use one letter or two letters or whatever. I find words to be more readable, predictable, and less confusing.

nathanstilwell: @Annebirzin @rsadres Regarding Icon sizes ☝️ Does design have a comprehensive list of sizes for icons? I'm not sure if I saw an audit done for icons.

packages/ui-components/src/Icon/README.md
nathanstilwell: I think since this is causing confusion it should be removed. I like the idea of having documentation for these components in the same repo as the code, but agree that the maintenance is a more important issue. Rachel and Anne are exploring plugins for storybook that I think would be the best place for all documentation including props, usage, and code examples. After I wrap this component, I'll circle back to cleaning up the READMEs and migrating example usage into storybook.

packages/ui-components/src/Icon/README.md
nathanstilwell: agreed. this was before we had tokens and the references to these colors was buried in Figma, which I thought was a big jump from the code.

PR #44, opened 2020-05-05
added npm scripts to simplify local setup and running dev servers

elkmaster:
# UI

Provided 2 new npm scripts on top level:
`npm run init` - will install all node modules and run needed lerna commands for local setup
`npm start` - will start start `watch` dev servers for all packages inside one terminal window. no need to cd manually and having multiple terminals.

#### Checklist
- [ ] ~I have written or updated test for the changes I made~
- [ ] ~I have added or updated Storybook if appropriate for my changes~

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,5 +3,9 @@
   "private": true,
   "devDependencies": {
     "lerna": "^3.20.2"
+  },
+  "scripts": {
+    "init": "yarn && lerna bootstrap --force-local && lerna link",
+    "start": "lerna run start --parallel"
   }
 }
--- a/packages/ui-components/package.json
+++ b/packages/ui-components/package.json
@@ -9,12 +9,12 @@
   "main": "dist/main.js",
   "scripts": {
     "build": "webpack --display-error-details",
+    "build:watch": "webpack --watch",
     "ci": "jest --ci -i",
     "clean": "rm -rf  ./dist/*",
     "lint": "eslint ./src/**/*.{tsx,ts,js} --format=codeframe",
     "test": "jest --watch",
-    "watch": "webpack",
-    "start": "npm run watch"
+    "start": "npm run build:watch"
   },
   "keywords": [],
   "license": "MIT",
--- a/packages/ui-components/package.json
+++ b/packages/ui-components/package.json
@@ -13,7 +13,8 @@
     "clean": "rm -rf  ./dist/*",
     "lint": "eslint ./src/**/*.{tsx,ts,js} --format=codeframe",
     "test": "jest --watch",
-    "watch": "webpack"
+    "watch": "webpack",
+    "start": "npm run watch"
   },
   "keywords": [],
   "license": "MIT",
--- a/README.md
+++ b/README.md
@@ -1,2 +1,7 @@
 # ui
 Published assets for Cockroach Labs user interfaces
+
+Commands for local setup:
+
+`npm run init` - for initial run, will install top level dependencies, package level ones, run lerna link
+`npm start` - run in root of the project to run dev servers for all sub packages.<br/>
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -3,5 +3,9 @@
   "private": true,
   "devDependencies": {
     "lerna": "^3.20.2"
+  },
+  "scripts": {
+    "init": "yarn && lerna bootstrap --force-local && lerna link",
```
laurenbarker, line 8: [`lerna bootstrap`](https://github.com/lerna/lerna/tree/master/commands/bootstrap) should install dependencies and symlink all of the packages so `yarn` and `lerna link` shouldn't be necessary.

laurenbarker: Why `--force-local`? I haven't worked with lerna very much so I'm not sure if this is common but I would think our versions should match when running `init`.

> When passed, this flag causes the bootstrap command to always symlink local dependencies regardless of matching version range.

packages/ui-components/package.json
```diff
@@ -13,7 +13,8 @@
     "clean": "rm -rf  ./dist/*",
     "lint": "eslint ./src/**/*.{tsx,ts,js} --format=codeframe",
     "test": "jest --watch",
-    "watch": "webpack"
+    "watch": "webpack",
```
laurenbarker, line 16: I know you didn't add this line but should it be `webpack --watch`?

laurenbarker: Maybe also unify with the icons package and change the name to `build:watch` since `start` will simplify it anyway.

package.json
elkmaster: first yarn is needed to install top level packages. `lerna` call will fail without it, except case when you have installed lerna -g.

`--force-local` is not mandatory but due to `^` version notation in packages we can catch an edge case when newer version of package being published to npm, but not merged in main brunch. or you in the middle of development of some feature and do not want or forget to rebase and need to run `init`. I am not sure in how realistic this scenarios. With `--force-local` you avoid that external package being fetched and\or cached somewhere, and you end up with making changes, and do not seeing them while hot-reload. 

additional lerna link is also not mandatory but, at least on my machine I encountered cases when I need it additionally after bootstrap. I don't know due to what it happening (npm caches, wrong moon phase?) but adding this as after bootstrap works pretty stable.

packages/ui-components/package.json
elkmaster: `webpack` command in ui-components starts webpack dev server. and `build:watch` seems to actually create dist files on every change. So I would not rename them as this commands already sorta correctly named to be self explanatory.

laurenbarker: We don't have `webpack-dev-server` installed but it looks like watch is set to true in the webpack config https://github.com/cockroachdb/ui/blob/master/packages/ui-components/webpack.config.js#L7

package.json
laurenbarker: Thanks for the explanation! Would you mind adding then instructions to get things started and the explanation of what the commands do to the README?

packages/ui-components/package.json
laurenbarker: I think removing `watch` from the config and then updating `"watch": "webpack"` to be `"build:watch": "webpack --watch"` would be the most clear here. It's building when files change like the `ui-components` repo. Does that make sense to you @elkmaster?

packages/ui-components/src/index.js
```diff
@@ -1 +1,2 @@
 export { Badge } from "./Badge";
+export { Tooltip } from "./Tooltip";
```
laurenbarker, line 2: Looks like the tooltip import snuck in here

Issue #22, opened 2020-04-08
Avatar Component


nathanstilwell:
# Avatar
> Cockroach Core Components Candidate

 _A temporary manifestation or aspect of a continuing entity._ Avatars are iconic representations of a user, organization, action, step in progression, or result.

![Screen Shot 2020-04-08 at 18 10 37](https://user-images.githubusercontent.com/397448/78838494-4e27bf00-79c4-11ea-8ee3-4c27e77bbaa4.png)

### Existing Components
- `UserAvatar` from [cockroachdb/cockroach](https://github.com/cockroachdb/cockroach/blob/27debe8006fc99a8e9db9bae6407796a2d6e6f23/pkg/ui/src/views/shared/components/userAvatar/index.tsx)
- `Avatar` from CockroachCloud

koorosh:
It fills like `Steps` and `Validation` components (shown on the attached screenshot) have to be stand-alone components because they have very little in common with Avatars.

nathanstilwell:
@koorosh I think that is true that `<Avatar>` shouldn't fully encapsulate the functionality of steps or validation, but I think those components could use `<Avatar>` as a dependency.

PR #48, opened 2020-05-06
ui-components: Avatar

koorosh:
# ui-components // Avatar

Issue #22

- Includes base implementation of <Avatar> component which includes
the ability to modify intent, size, disable, and set selectable props
to component.
- It is a starting point to extend this implementation for more
custom representations as StepAvatar and Validation which require
customization of styles.
- Storybook is updated with one more story to show <Avatar> component in different
states, it is a draft version to validate current implementation only.

#### Checklist

### DIFF
--- a/packages/ui-components/src/Avatar/Avatar.module.scss
+++ b/packages/ui-components/src/Avatar/Avatar.module.scss
@@ -41,32 +41,18 @@
   color: $crl-red-4;
 }
 
-.size-l {
+.size-default {
   width: 40px;
   height: 40px;
   font-size: 14px;
   letter-spacing: 0.1px;
 }
 
-.size-m {
+.size-small {
   width: 32px;
   height: 32px;
   font-size: 12px;
   letter-spacing: 0.3px;
-}
-
-.size-s {
-  width: 24px;
-  height: 24px;
-  font-size: 14px;
-  letter-spacing: 0.1px;
-}
-
-.size-xs {
-  width: 16px;
-  height: 16px;
-  font-size: 10px;
-  letter-spacing: 0.1px;
 }
 
 .noTransform {
--- a/packages/ui-components/src/Avatar/Avatar.module.scss
+++ b/packages/ui-components/src/Avatar/Avatar.module.scss
@@ -1,0 +1,99 @@
+@import "../styles/tokens.scss";
+
+// TODO (koorosh): Probably has to be extracted to tokens.
+@mixin selectable-border($color) {
+  box-sizing: border-box;
+  border: 2px solid $color;
+}
+
+.avatar {
+  position: relative;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  flex-shrink: 0;
+  font-family: Source Sans Pro, sans-serif; // TODO (koorosh): replace with fonts from tokes
+  font-style: normal;
+  font-weight: 600;
+  line-height: 1;
+  border-radius: 50%;
+  overflow: hidden;
+  user-select: none;
+}
+
+.intent-default {
+  background: $crl-neutral-2;
+  color: $crl-neutral-7;
+}
+
+.intent-active {
+  background: $crl-blue-1;
+  color: $crl-base-blue;
+}
+
+.intent-pending {
+  background: $crl-orange-1;
+  color: $crl-orange-4;
+}
+
+.intent-invalid {
+  background: $crl-red-1;
+  color: $crl-red-4;
+}
+
+.size-l {
+  width: 40px;
+  height: 40px;
+  font-size: 14px;
+  letter-spacing: 0.1px;
+}
+
+.size-m {
+  width: 32px;
+  height: 32px;
+  font-size: 12px;
+  letter-spacing: 0.3px;
+}
+
+.size-s {
+  width: 24px;
+  height: 24px;
+  font-size: 14px;
+  letter-spacing: 0.1px;
+}
+
+.size-xs {
+  width: 16px;
+  height: 16px;
+  font-size: 10px;
+  letter-spacing: 0.1px;
+}
+
+.disabled {
+  background: $crl-neutral-2;
+  color: $crl-base-text--light;
+
+  &:focus, &:hover {
+    border: none;
+  }
+}
+
+.selectable:not(.disabled) {
+  &:focus, &:hover {
+    &.intent-default {
+      @include selectable-border($crl-neutral-7);
+    }
+
+    &.intent-active {
+      @include selectable-border($crl-base-blue);
+    }
+
+    &.intent-pending {
+      @include selectable-border($crl-orange-4);
+    }
+
+    &.intent-invalid {
+      @include selectable-border($crl-red-4);
+    }
+  }
+}
--- a/packages/ui-components/src/Avatar/Avatar.module.scss
+++ b/packages/ui-components/src/Avatar/Avatar.module.scss
@@ -69,6 +69,18 @@
   letter-spacing: 0.1px;
 }
 
+.noTransform {
+  text-transform: none;
+}
+
+.transformCase-none {
+  text-transform: none;
+}
+
+.transformCase-uppercase {
+  text-transform: uppercase;
+}
+
 .disabled {
   background: $crl-neutral-2;
   color: $crl-base-text--light;
--- a/packages/ui-components/src/Avatar/Avatar.test.tsx
+++ b/packages/ui-components/src/Avatar/Avatar.test.tsx
@@ -9,7 +9,7 @@ describe("Avatar", () => {
       const wrapper = shallow(<Avatar>CL</Avatar>);
       const className = wrapper.prop("className");
       expect(className).toContain("intent-default");
-      expect(className).toContain("size-l");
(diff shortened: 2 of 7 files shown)


packages/ui-components/src/Avatar/Avatar.module.scss
```diff
@@ -0,0 +1,98 @@
+@import "../styles/tokens.scss";
+
+@mixin selectable-border($n) {
+  box-sizing: border-box;
+  border: 2px solid $n;
+}
```
j-low, line 6: This seems like it would be a good candidate to be added to our tokens, probably with a default color value.  It might not be the right time to do so now (i.e. when we have more use cases) but let's add a TODO comment above this to call it out.

Also the let's make var name something descriptive; for example `$color`... `$n` has connotations with mathematics as relating to an integer count value so it makes it slightly (though not fatally) confusing.

packages/ui-components/src/Avatar/Avatar.module.scss
```diff
@@ -0,0 +1,98 @@
...
+
+.intent-default {
+  background: $crl-neutral-2;
+  color: $crl-neutral-7;
+}
+
+.intent-info {
+  background: $crl-blue-1;
+  color: $crl-blue-3;
```
j-low, line 30: Wherever possible (or unless there's a specific reason not to do so) let's refer to the `base` aliases, in this case `$crl-base-blue`.  The reason being the `base` aliases will be the most common colors seen in the codebase and, conversely, the explicit numerical values should be uncommon enough to call themselves out as a idiosyncratic use case when they do show up.

packages/ui-components/src/Avatar/Avatar.module.scss
```diff
@@ -0,0 +1,98 @@
...
+}
+
+.intent-warning {
+  background: $crl-yellow-1;
+  color: $crl-yellow-4;
```
j-low, line 35: Our warning color is `$crl-base-yellow`/`$crl-yellow-3`; comment re: base colors from above applies here too.

packages/ui-components/src/Avatar/Avatar.module.scss
```diff
@@ -0,0 +1,98 @@
...
+}
+
+.intent-danger {
+  background: $crl-red-1;
+  color: $crl-red-4;
```
j-low, line 40: Our warning color is `$crl-base-red`/`$crl-red-3`; comment re: base colors from above applies here too.

packages/ui-components/src/Avatar/Avatar.module.scss
```diff
@@ -0,0 +1,98 @@
...
+}
+
+.selectable:not(.disabled) {
+  &:focus, &:hover {
+    &.intent-default {
+      @include selectable-border($crl-neutral-7);
+    }
+
+    &.intent-info {
+      @include selectable-border($crl-blue-3);
```
j-low, line 87: Comment re: base colors from above applies here too.

packages/ui-components/src/Avatar/Avatar.module.scss
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    &.intent-warning {
+      @include selectable-border($crl-yellow-4);
```
j-low, line 91: Our warning color is `$crl-base-yellow`/`$crl-yellow-3`; comment re: base colors from above applies here too.

packages/ui-components/src/Avatar/Avatar.module.scss
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    &.intent-danger {
+      @include selectable-border($crl-red-4);
```
j-low, line 95: Our warning color is `$crl-base-red`/`$crl-red-3`; comment re: base colors from above applies here too.

packages/ui-components/src/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,59 @@
...
+    </div>
+  );
+};
+
+Avatar.defaultProps = {
+  size: "40",
+  intent: "default",
+  selectable: false,
+  disabled: false,
+};
```
j-low, line 57: As far as setting default props, I kind of like this pattern here but I know in other places we've specified them in function signature itself. @laurenbarker do you have an opinion on this at all?

packages/ui-components/src/Avatar/Avatar.test.tsx
```diff
@@ -0,0 +1,89 @@
+import React from "react";
+import { shallow } from "enzyme";
```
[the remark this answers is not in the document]
laurenbarker: Was thinking the same thing! Created one this morning https://github.com/cockroachdb/ui/issues/49

packages/ui-components/src/Avatar/Avatar.tsx
laurenbarker: `defaultProps` is deprecated for function components (https://github.com/facebook/react/pull/16210) which is why we decided to go with [default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) elsewhere.

packages/ui-components/src/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,59 @@
...
+export interface AvatarProps {
+  children?: string;
+  size?: AvatarSize;
+  intent?: AvatarIntent;
+  disabled?: boolean;
+  selectable?: boolean;
+  onClick?: () => void;
+}
+
+export type AvatarSize = "16" | "24" | "32" | "40";
```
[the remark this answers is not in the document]
koorosh: Agree, 👍🏽 the only thing, I would start from "xs" as the smallest one for 16px.

packages/ui-components/src/Avatar/Avatar.module.scss
koorosh: Gotcha!

packages/ui-components/src/Avatar/Avatar.module.scss
koorosh: Actually, according to the designs in [Figma](https://www.figma.com/file/B5AtEGdRbW3VxiBdg62TB8Xw/CRDB-Design-System_V2?node-id=72%3A0)  it is supposed to be `yellow-1` and `yellow-4` colors.

Probably the naming for `warning` and `danger` intents are misleading and have to be used the same as in designs: `invalid` and `pending`? What do you think?

j-low: Okay, my apologies, It looks like there is some divergence from the Figma DS and what was encoded as tokens in the cloud repo at an earlier date (which implies I'll have to do some refactoring of that repo's tokens instead of vice versa).

That said, it looks like the warning colors are orange instead of yellow. 

Agreed that the intent labels would be better as `invalid` and `pending` as they reflect avatar-related states more accurately from a semantic standpoint.

koorosh: Done, names of intents are refactored to reflect the same names as in designs.

packages/ui-components/src/Avatar/Avatar.module.scss
koorosh: > Just the one issue re: orange vs. yellow and updating intent labels but otherwise looks great
Updated PR with this change as well! Thank you for the comment!

koorosh:
> Just the one issue re: orange vs. yellow and updating intent labels but otherwise looks great

Fixed!

Annebirzin:
@koorosh Rachel and I were able to review the avatar component and had some feedback:

- force the text to be all-caps (no lowercase) 
- the font weight 600 doesn't look bold as it does in Figma, is there an issue there?
- Remove the `Small` avatar size (24 x24). We decided the use case for this size outlined in [avatars & badges](https://www.figma.com/file/B5AtEGdRbW3VxiBdg62TB8Xw/CRDB-Design-System_V2?node-id=72%3A0) section of  design system is not an avatar but more of a step component like seen in [ant design](https://ant.design/components/steps/). Will update the component in Figma to remove the 24x24 and 16x16 size.

koorosh:
> @koorosh Rachel and I were able to review the avatar component and had some feedback:
> 
> * force the text to be all-caps (no lowercase)
> * the font weight 600 doesn't look bold as it does in Figma, is there an issue there?
> * Remove the `Small` avatar size (24 x24). We decided the use case for this size outlined in [avatars & badges](https://www.figma.com/file/B5AtEGdRbW3VxiBdg62TB8Xw/CRDB-Design-System_V2?node-id=72%3A0) section of  design system is not an avatar but more of a step component like seen in [ant design](https://ant.design/components/steps/). Will update the component in Figma to remove the 24x24 and 16x16 size.

@Annebirzin , thanks for your feedback!
Following changes were made:
- By default, text transformed to upper-case. It uses the same logic as `Badge` component where user still have an option to explicitly indicate to keep text with original casing.
- Double checked font weights, it is 600 and has all the same styles as Figma designs.
- Only two sizes are left (40px and 32px)

Annebirzin:
thanks @koorosh!

PR #50, opened 2020-05-06
[storybook] TS integration

### LINKED ISSUE #45: [storybook] add Typescript support
Currently stories can only be written in pure JS which prevents us from gaining the benefits of type safety.

Info: https://storybook.js.org/docs/configurations/typescript-config/

j-low:
# storybook-ui-components // TS integration

* resolves #45 
* the ambition of this PR has been scoped-down in the interest of keeping work moving and not blocking work on components (which is presently being done in Typescript).  I've left a [fairly extensive comment](https://github.com/cockroachdb/ui/pull/50/files#diff-98397f3684de78665d1cc2ba50c33075R17-R55) summarizing the scope-cut and have indicated next steps for follow-up.
* the upshot of the above is that until #53 is resolved we will have to be on a CR-enforced "honor" system w/r/t to leveraging the type system appropriately.

#### Checklist
- ~[ ] I have written or updated test for the changes I made~
- ~[ ] I have updated the README of the package I'm working in to reflect my changes~


packages/storybook-ui-components/tsconfig.json
```diff
@@ -0,0 +1,66 @@
...
+
+      ```
+      Could not find a declaration file for module '@cockroachlabs/ui-components'
+      ...implicitly has an 'any' type.
+
+      Try `npm install @types/cockroachlabs__ui-components` if it exists or add
+      a new declaration (.d.ts) file containing `declare module '@cockroachlabs/ui-components';`
+      ```
+
+      To solve this is issue seems to require reconfiguring Webpack and Typescript
```
laurenbarker, line 32: After talking to someone who works on typescript library, I don't think we need for the types to be in one file. The second option in https://medium.com/@NiGhTTraX/how-to-set-up-a-typescript-monorepo-with-lerna-c6acda7d4559 (using `composite: true`) is what they do for their company's shared library. Unfortunately that's not open sourced but I think the article provides a good overview.

j-low:
> I don't think having one declaration file is necessary though. I can try what is suggested in the article I linked in the comments tomorrow to see is a workable solution.

That sounds good. Merging this should prevent component PRs from piling up but I'm happy to work on fixing this setup next week.  A single `tsconfig.json` would be great too.

Issue #36, opened 2020-04-23
[tracking] READMEs as documentation


nathanstilwell:
We have a couple of READMEs in the repo, but I think at a minimum we should have one at the root of every package.

laurenbarker:
https://github.com/cockroachdb/ui/pull/44 adds development commands to the main README and in the comments there is some more context around those. We should probably define what goes in the README.md vs `CONTRIBUTING.md` though because it's unclear to me. cc @j-low

laurenbarker:
Following up on defining what goes in each doc:

My confusion about what goes into our `CONTRIBUTING.md` vs `README.md` is because our installation and publishing instructions are in `CONTRIBUTING.md`. Those instructions are typically in the `README.md` so that's where I would look for them. Also, we put build/installation instructions for each package in that package's `README.md` so it is kind of inconsistent.

After looking into this a bit more, I think we should move the installation and publishing information to the `README.md`. Right now I think we are mixing [contribution guidelines](https://help.github.com/en/github/building-a-strong-community/setting-guidelines-for-repository-contributors#examples-of-contribution-guidelines) and getting started instructions.

nathanstilwell:
circling back to this based on conversations we been having recently. A couple of things,

* `README` should define usage and user context about the package they are located in.
* `CONTRIBUTING` should contain information about developing inside of the UI repo. We also recently made the decision to have only one top level `CONTRIBUTING` file. Where there are differences in development among the packages, that complexity should be obfuscated behind `npm` lifecycle scripts. 

There may be details that don't fall into either of these categories, but in general if it is for users of UI libraries it goes in a README and if it is for developers of UI libraries it goes in CONTRIBUTING.

PR #51, opened 2020-05-07
Tooltip

elkmaster:
# ui-components // tooltip

Used popper.js library for positioning.

There is a story books available to have a quick look on position support.

Screenshot 2020-05-08 at 18 54 16

#### Checklist


packages/ui-components/src/Tooltip/Tooltip.scss
```diff
@@ -0,0 +1,189 @@
...
+$tooltip-bg: #ffffff;
+$tooltip-opacity: 1;
+
+$tooltip-border-width: 1px;
+$tooltip-border-color: #b1b1b1;
+$tooltip-shadow-width: 1px;
+
+$tooltip-arrow-width: 6px;
+$tooltip-arrow-color: $tooltip-border-color;
+$tooltip-arrow-inner-color: $tooltip-bg;
```
j-low, line 24: Font sizes, colors, border radii, shadows, etc. should definitely be tokenized and consumed as opposed to defined here.  Since we only have a small number of things tokenized right now, I'd recommend subbing in token values wherever possible, adding a TODO above these styles re: tokenization, and filing an issue to create tokens for anything that might be missing in tokens that is being defined here or elsewhere.

packages/ui-components/src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,47 @@
...
+  ...props
+}) => {
+  if (!content) {
+    return null;
+  }
+
+  return (
+    <RCtooltip
+      {...props}
+      prefixCls={"tooltip"}
```
j-low, line 34: Let's set the base prefix for components `.crl-my-component-name` for clarity

packages/ui-components/src/Tooltip/Tooltip.scss
```diff
@@ -0,0 +1,218 @@
...
+  }
+  .crl-tooltip-arrow-inner {
+    border-color: $bg;
+  }
+}
+
+//
+// Default style
+// --------------------------------------------------
+.default {
```
laurenbarker, line 29: Since this isn't using css modules these styles will be global and could also cause conflicts in consuming apps.

elkmaster: `:global` namespace not working good with current styles due to a bit tricky way of styling tooltip arrow and provide easy to adjust theme classes. but I moved that definitions inside `.crl-tooltip` namespace to minify possible conflicts

package.json
```diff
@@ -3,5 +3,9 @@
   "private": true,
   "devDependencies": {
     "lerna": "^3.20.2"
+  },
+  "scripts": {
+    "init": "yarn && lerna bootstrap --force-local && lerna link",
```
nathanstilwell, line 8: I believe that `lerna bootstrap` performs a dependency symlinking. Do we need an additional`lerna link` here?

packages/ui-components/src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,53 @@
...
+  | "topRight"
+  | "bottomLeft"
+  | "bottomRight";
+
+type TooltipStyle = "default" | "light" | "dark";
+
+interface TooltipProps {
+  position?: TooltipPosition;
+  style?: TooltipStyle;
+  content: ReactElement | string;
```
nathanstilwell, line 21: In developing the `Badge` component, there was some discussion that arose from [a comment](https://github.com/cockroachdb/ui/pull/3#discussion_r382102378) about a `content` prop vs `children`. We landed on using only `children` which makes the API for the component more intuitive. I was tempted at first to use a `content` prop also at first as it could allow for tighter control, but ultimately saw the wisdom is the flexibility of allowing the user to use child markup.

By removing the `content` prop it will also reduce the complexity of the component slightly as the code fork below in checking for content won't be required.

packages/ui-components/src/Tooltip/Tooltip.test.tsx
```diff
@@ -0,0 +1,40 @@
+import React from "react";
+import { shallow, mount } from "enzyme";
+
+import Tooltip from "./Tooltip";
```
nathanstilwell, line 4: I'll echo Lauren's concern above, that we will need to use CSS Modules here as I think it's important to keep these components completely encapsulated. I am opposed to having any core component rely on any global style. I am checking out your branch to see if I can help investigate the challenges you were having with CSS Modules.

package.json
laurenbarker: @nathanstilwell the first two commits are actually already in master. I asked the same question https://github.com/cockroachdb/ui/pull/44#discussion_r420134062

Also added a note as part of the README ticket in case we want to adjust anything https://github.com/cockroachdb/ui/issues/36#issuecomment-624882442

nathanstilwell: Thanks for the head's up @laurenbarker.

packages/ui-components/src/Tooltip/Tooltip.tsx
elkmaster: we already use children, see an example:
`<Tooltip content={"tooltip text"}>
       <div>hover me</div>
 </Tooltip>`

nathanstilwell: My fault. I just got the API of this component. `children` is the _target element_ of the Tooltip and the `content` is the content of the tooltip itself.

packages/ui-components/src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,102 @@
...
+      <div
+        className={classnames}
+        ref={setPopperElement}
+        style={styles.popper}
+        data-jest="tooltip"
+        {...attributes.popper}
+      >
+        {content}
+        <div className={css.arrow} ref={setArrowElement} style={styles.arrow} />
+        <div className={css.arrowInner} style={styles.arrow} />
```
nathanstilwell, line 96: Tell me about why we need this `arrowInner`. I'm playing around with it in the browser and I can tell what difference it makes if it's there or not.

elkmaster: this needed if border of tooltip differs in color from background. for example, here is a `light` style without `arrowInner`:
Screenshot 2020-05-27 at 13 46 49

Issue #23, opened 2020-04-08
Input Component


nathanstilwell:
# Input
> Cockroach Core Components Candidate

A control allowing the user to enter text by typing or interacting with other controls. This component should be concerned with extensibility over state management.

![Screen Shot 2020-04-08 at 18 37 52](https://user-images.githubusercontent.com/397448/78840176-13278a80-79c8-11ea-80db-a22a7485fe03.png)

_**There's kind of a lot going on here, but I think this component needs a name to suggest it will support composition into other controls. This could serve as the base for text input, select menus, password or email input fields, etc_

### Existing Components
- [`TextInput`](https://github.com/cockroachdb/cockroach/blob/27debe8006fc99a8e9db9bae6407796a2d6e6f23/pkg/ui/src/components/textInput/index.tsx), [`Search`](https://github.com/cockroachdb/cockroach/blob/27debe8006fc99a8e9db9bae6407796a2d6e6f23/pkg/ui/src/views/app/components/Search/index.tsx) from [cockroachdb/cockroach](https://github.com/cockroachdb/cockroach)
- `InputField`, `TextInput`, `ValidatedInput` and `Input` from CockroachCloud

Annebirzin:
For reference, adding a Figma link of the Input audit which demonstrates examples of where inputs appear across CockroachCloud and Admin UI:

[Input Audit](https://www.figma.com/file/TqjqzNz8WpGJCTy0UBFxHB/obsrv-design-components-v1?node-id=1359%3A3177)

PR #52, opened 2020-05-07
ui-components: Input

koorosh:
# ui-components // Input

Resolves: #23 
Base implementation of `Input` component
Added Storybook as a playground to validate styles.
Supports only `text` type yet.

- `<TextInput />`
- `<NumberInput />`
- `<BaseInput />`

#### Checklist


packages/ui-components/src/Input/NumberInput.tsx
```diff
@@ -0,0 +1,75 @@
...
+export type NumberInputProps = Omit<InputProps<number>, "type">;
+
+export const NumberInput: React.FC<NumberInputProps> = ({
+  onChange,
+  value: outerValue,
+  initialValue,
+  prefix,
+  ...props
+}) => {
+  const { invalid, disabled } = props;
```
nathanstilwell, line 20: Why  destructure  these here and not up in the parameter list with the other?

packages/ui-components/src/Input/number-input.scss
```diff
@@ -0,0 +1,76 @@
...
+.input[type="number"]::-webkit-outer-spin-button,
+.input[type="number"]::-webkit-inner-spin-button {
+  -webkit-appearance: none;
+  margin: 0;
+  pointer-events: none;
+}
+
+// remove default arrow spinner for input -- Firefox browser
+.input[type=number] {
+  -moz-appearance: textfield;
```
nathanstilwell, line 14: In Firefox, the spin buttons are being pushed out of the width of the input container by the `width: 100%` of the base input.

Screen Shot 2020-05-21 at 12 29 33

```suggestion
  -moz-appearance: textfield;
  padding-right: 0;
  width: calc(100% - 32px);
```

packages/ui-components/src/Input/input.scss
```diff
@@ -0,0 +1,92 @@
...
+  }
+}
+
+// Prefix component styles. Set constant width/height
+// and apply colors based on container's state.
+.container {
+  .prefix {
+    width: 14px; // TODO (koorosh): has to be the same as font-size
+    min-width: 14px;
+    height: 14px;
```
nathanstilwell, line 82: In Firefox, the prefix icon is misaligned

Screen Shot 2020-05-21 at 12 38 22

```suggestion
    height: 22px;
    line-height: 22px;
```

packages/ui-components/src/Input/BaseInput.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const [value, setValue] = useState<string | number>(
+    outerValue || initialValue,
+  );
+  const [isDirty, setDirtyState] = useState<boolean>(false);
+
+  useEffect(() => {
+    if (!outerValue && !isDirty) {
+      return;
+    }
+    setDirtyState(true);
```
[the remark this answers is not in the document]
koorosh: @nathanstilwell , you're right it's useless duplication in `onChangeHandler`. Will remove it.
```
setDirtyState(true);
setValue(nextValue);
```

The current implementation is necessary to achieve the following conditions:
- `defaultValue` is shown only when `value` isn't set and user didn't change the text
- New `value` provided from outside of the component overrides `defaultValue` and the current value of Input
- Typed value by a user should not be immediately overridden by the outer `value` prop.

Basically it allows us to change input value in two ways: programmatically by changing `value` prop or by user input.
To manage two-way changes for input - it's necessary to maintain an internal state for value. So `value` props is not directly passed to <input> `value` attribute. Instead, change of `value` props (`outerValue`) is tracked in `useEffect`.

`isDirty` flag is required to handle one corner case - when `value` prop is set to empty string or user removes input text. We need to distinguish whether to show `initialValue` or not. After value is remove - it is dirty state and no needs to show initial value.

packages/ui-components/src/Input/BaseInput.tsx
```diff
@@ -0,0 +1,94 @@
...
+  autoComplete = "off",
+  className,
+  style,
+  value: outerValue = "",
+  initialValue = "",
+  placeholder,
+  onChange,
+  disabled = false,
+  invalid = false,
+}) => {
```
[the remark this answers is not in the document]
koorosh: Currently, `InputProps` doesn't allow to provide any unspecified props, so there is nothing left in props and the user cannot add anything else.
Do you suggest that we need to allow users to define all possible Input values, even if we don't handle them inside of component? For instance, `onFocus` or `onKeyDown` events?

packages/ui-components/src/Input/NumberInput.tsx
koorosh: Good point! The reason was to pass rested `props` to `<BaseInput {...props} ... />` and have these props inside rested object. But it already has other props defined, so will destruct it in the parameters list. Thanks!

packages/ui-components/src/Input/number-input.scss
koorosh: Wow, missed testing for `NumberInput` on Firefox.
Found that `overflow-x` attribute fixes this case. Moreover, `NumberInput` might have or not the `prefix` icon so calculating width would be much complicated.

Also added `firefox-only` mixing to apply styles for Firefox only.

packages/ui-components/src/Input/BaseInput.tsx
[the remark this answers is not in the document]
nathanstilwell: I’m a little torn between allowing unhandled props or specifying every prop possible. As you said, I don’t want to prevent users from adding additional event handlers or accessibility props or data props or any valid prop of an HTML input. But in a previous implementation I found it unruly to specify every possible prop when the base case of the component will only use a small portion of the props.

Having said that, I get the feeling that passing unhandled props are anathema to type safety. But I suppose my answer to that would be that HTML isn’t type safe. I welcome further discussion if you have suggestions.

packages/ui-components/src/Input/input.scss
koorosh: @nathanstilwell , I couldn't reproduce this issue. 
for this element, `margin-top` and `margin-bottom` are set to auto - so it has to do a trick and center element.
`height` and `width` define the size of SVG icon and it has to be the same as font size.
Please, let me know if there are some specific conditions required to reproduce this.

packages/ui-components/src/Input/BaseInput.tsx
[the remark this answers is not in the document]
nathanstilwell: Thank you for the description. I was concerned that I was missing something with the duplication.

I agree with your approach to handling a little internal state. I’m wonder how this will work with something like React Final Form, which is our proposed form solution in Cloud console (and I would like to use it to manage all forms if possible). My gut feeling is that it will be fine as the Field props will be passed down and we will probably still have cases where we want to us an Input without using a Form or Field from RFF.

packages/ui-components/src/Input/number-input.scss
nathanstilwell: Perfect, I didn’t consider the prefix for width calculation.

packages/ui-components/src/Input/input.scss
nathanstilwell: Interesting. I’m using FF developer edition, but if the issue isn’t present in the stable version of FF, I wouldn’t worry about it.

I do have a concern that all of the icons are not square and so wonder if that will effect how they are rendered. This is also an issue I would like to address from the Icon side as having square icons that are all the same base size will make this kind of thing simpler. @Annebirzin

packages/ui-components/src/Input/BaseInput.tsx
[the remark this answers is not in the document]
koorosh: We could extend props with `React.InputHTMLAttributes<HTMLInputElement>` which is React type for allowed `<input />` props.
- we preserve type safety
- restrict users to provide only valid props

packages/ui-components/src/Input/BaseInput.tsx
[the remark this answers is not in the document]
koorosh: Well, it's a great example to add in Storybook - to check out how this our component work with RFF! 👍🏽

packages/ui-components/src/Input/input.scss
Annebirzin: @nathanstilwell yep agree with making all icons square. I can work with you on this

packages/ui-components/src/Input/BaseInput.tsx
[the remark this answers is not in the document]
nathanstilwell: That’s perfect.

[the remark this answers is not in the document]
koorosh: Done

packages/ui-components/src/Input/TextInput.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { BaseInput, InputProps } from "./BaseInput";
+import { InputPrefix, InputWrapper } from "./helpers";
+
+export type TextInputProps = Omit<InputProps<string>, "type">;
+
+export const TextInput: React.FC<TextInputProps> = props => {
+  const { prefixIcon, disabled, invalid } = props;
+  return (
+    <InputWrapper disabled={disabled} invalid={invalid}>
+      <InputPrefix>{prefixIcon}</InputPrefix>
```
nathanstilwell, line 11: I don't want to hold up this PR any longer, but wanted to leave a note to come back to. Should we conditionally render this prefixIcon only if an icon is provided? Will React be smart enough to not render empty DOM elements?

Issue #27, opened 2020-04-20
[Tracking] Design Tokens


laurenbarker:
## What
Add design-tokens package to the UI Repo.

## Why
Ensure brand styles (colors, spacing, typography, etc.) can be included in all of our user facing apps while maintaining a single source of truth.

## Phase 1
- [ ] Add design-tokens package to the UI Repo
- [ ] Manually add our brand styles in JSON format in a similar format to base.json in the prototype (link in design doc)
    * To begin with, the brand styles to include are the variables currently defined in CC
    * Each category of token (color, grid, etc.) will require a different level of organization. Wunderzin’s base.json provides some examples of tokens that have many variations.
- [ ] Configure Style Dictionary to output our brand styles in three formats: SASS, JS, Stylus
    * SCSS and JS are [pre-defined style dictionary formats](https://amzn.github.io/style-dictionary/#/formats?id=pre-defined-formats) but Stylus would require a [custom formatter](https://amzn.github.io/style-dictionary/#/formats?id=creating-formats)
    * The high level categories can be split into separate files. Wunderzin’s [web build](https://github.com/PavelLaptev/wunderzin/tree/master/style-dictionary/build/web) provides an example.
- [ ] Publish the design-tokens package to npm
- [ ] Consume the npm package in CC, shared component library, and wherever else the brand styles are needed

## Phase 2
- [ ] generate token from Figma (see design doc)

PR #54, opened 2020-05-13
[design-tokens] Design token initial - colors

laurenbarker:
# design-tokens // add design token package
Related to https://github.com/cockroachdb/ui/issues/27

Uses [style-dictionary](https://github.com/amzn/style-dictionary) to transform tokens defined in json into SCSS and JS variables.

See https://github.com/cockroachdb/ui/pull/62 for usage.

**Example output:**
```scss
// Do not edit directly
// Generated on Fri, 29 May 2020 22:15:25 GMT

$color-background-base: #ffffff;
$color-background-alt: #f5f7fa;
$color-background-inverse: #475872;
$color-background-success: #37a806;
$color-background-error: #ff3b4e;
$color-background-warning: #ffa53b;
$color-background-info: #0788ff;
$color-background-danger: #ff3b4e;
$color-background-badge-neutral: #e7ecf3;
$color-background-badge-success: #b4f1aa;
$color-background-badge-warning: #fef1da;
$color-background-badge-info: #e0eefb;
$color-background-badge-danger: #ffe5e8;
...
```

#### Checklist
- [ ] ~~I have written or updated test for the changes I made~~
- [ ] ~~I have added or updated Storybook if appropriate for my changes~~


packages/design-tokens/tokens/color/background.json
```diff
@@ -0,0 +1,39 @@
+{
+  "color": {
+    "background": {
+      "base": { "value": "{color.base.white.value}" },
+      "alt": { "value": "{color.core.neutral.1.value}" },
```
laurenbarker, line 5: This is the gray background we use some places. Not sure what to call it.

packages/design-tokens/tokens/color/background.json
```diff
@@ -0,0 +1,39 @@
...
+      "inverse": { "value": "{color.core.neutral.6.value}" },
+      "success": { "value": "{color.base.green.value}" },
+      "error": { "value": "{color.base.red.value}" },
+      "warning": { "value": "{color.base.orange.value}" },
+      "info": { "value": "{color.base.blue.value}" },
+
+      "button": {
+        "primary": {
+          "base": { "value": "{color.base.green.value}" },
+          "hover": { "value": "{color.base.green.value}" },
```
laurenbarker, line 15: We do a lot in the css right now to adjust the color:
```scss
    &:hover {
      background-color: $crl-green-4;
      color: $crl-base-white;
    }

    &:active {
      background-color: $crl-green-4;
      border-width: 1px;
      border-color: mix(black, $crl-green-4, 10%);
    }

    &:focus {
      border-width: 1px;
      border-color: mix(black, $crl-green-4, 20%);
      @include crl-box-shadow(0, 0, 0, 3px, rgba($crl-base-green, 0.4));
    }

    &:disabled {
      background-color: $crl-green-2;
      border-color: $crl-green-2;

      &:hover {
        cursor: not-allowed;
      }
    }
```

packages/design-tokens/tokens/color/font.json
```diff
@@ -0,0 +1,72 @@
+{
+  "color": {
+    "font": {
+      "1": { "value": "{color.core.neutral.8.value}" },
```
laurenbarker, line 4: These are the color variables used in figma

packages/design-tokens/tokens/color/background.json
nathanstilwell: I wonder if we could add an additional layer to the states with separate colors. For example for `hover`,
```json
...
"hover": {
  "background": { "value": "{color.base.green.value}" },
  "foreground": { "value": "{color.base.white.value}" },
},
```

I think should also move ahead and iterate. There may be some way to capture what Rachel had mentioned this morning about a number value representing a percent _darker_ or whathaveyou.

laurenbarker: Right now the text color is separate from the background. For example, the button text color on hover is [here](https://github.com/cockroachdb/ui/pull/54/files#diff-55f5b9d8b713974d4b9cc2eb1deb9af0R43). Is that what you mean by foreground?

PR #55, opened 2020-05-13
ui-components: Button component

### LINKED ISSUE #21: Button Component
# Button
> Cockroach Core Components Candidate

A UI element that a user can select to perform an action. While buttons can have _intents_ to express intention, this component should be purely [_presentational_](https://redux.js.org/basics/usage-with-react/#presentational-and-container-components) and contain no state or perform no action of its own. This component will be composed into other components and so should remain completely naive.
 
Screen Shot 2020-04-08 at 18 21 55

### `Intent`

From designs it seems that we have three _intents_, *Primary*, *Secondary*, and *Flat*.
Screen Shot 2020-04-23 at 17 34 40
Screen Shot 2020-04-23 at 17 34 44
Screen Shot 2020-04-23 at 17 34 48

### `Tint`

Inside of these _intents_, we occasionally have apply a color (or tint) to augment the meaning of the button, but this doesn't change it's intention. Some examples,

Screen Shot 2020-04-23 at 17 36 48
Screen Shot 2020-04-23 at 17 37 01
Screen Shot 2020-04-23 at 17 37 10

💭 _I'm not sure if all of these are use cases, but the "danger" or "red" tint is certainly a use case that exists and I can image other cases where we may employ standard colors to alter primary or secondary buttons._

### `Size`

There are two sizes defined that seemed be used pretty widely, "default" and "small".

Screen Shot 2020-04-23 at 17 44 06

This only seems to apply to height. In terms of width ...

### Fluid Width

Button widths seem to fall into two categories, _width based on content_ and _width based on container_.

*Based on content*
Screen Shot 2020-04-23 at 17 36 48
Screen Shot 2020-04-23 at 17 47 02
Screen Shot 2020-04-23 at 17 47 09
Screen Shot 2020-04-23 at 17 47 51

*Based on container*
Screen Shot 2020-04-23 at 17 48 58
Screen Shot 2020-04-23 at 17 49 03
Screen Shot 2020-04-23 at 17 49 34
Screen Shot 2020-04-23 at 17 50 15

💭 ❓ _My initial thought was "Fluid" as it's a prop I've used before as a boolean prop, which is not my favorite. I'm open to ideas here for a prop names that we could use for two distinct values. Either "default" and "fluid" or "content" and "container"._

### Icons

Buttons use icons in a few different ways. To highlight content or communicate functionality by having an icon to the right or the left of the text content,

Screen Shot 2020-04-23 at 18 26 05

or as the content,

Screen Shot 2020-04-23 at 18 30 26
Screen Shot 2020-04-23 at 18 30 31
Screen Shot 2020-04-23 at 18 30 56

💭 ❓ _One thing we may need council from design on, is if the icons are always part of the content or if they will have a layout different from the content. In the first case we may not need to include knowledge of icons at all in the implementation of button and could get away with `<Button><Icon /> Content</Button>`. If the icon needs alignment apart from the content (as seen often when buttons are used as trigger for menus) we have need something like `<Button iconLeft={Icon} iconRight={Icon}>Content</Button>`. Even in this case an "IconButton" would just not use these props and provide an Icon as content, e.g. `<Button>{Icon}</Button>`_

#### `Button` from [AntD](https://ant.design/components/button/)

```jsx
export interface BaseButtonProps {
  type?: ButtonType;
  icon?: React.ReactNode;
  shape?: ButtonShape;
  size?: SizeType;
  loading?: boolean | { delay?: number };
  prefixCls?: string;
  className?: string;
  ghost?: boolean;
  danger?: boolean;
  block?: boolean;
  children?: React.ReactNode;
}
```

#### `Button` from `[cockroachdb/cockroach](https://github.com/cockroachdb/cockroach/blob/27debe8006fc99a8e9db9bae6407796a2d6e6f23/pkg/ui/src/components/button/button.tsx)

```jsx
export interface ButtonProps {
  type?: "primary" | "secondary" | "flat";
  disabled?: boolean;
  size?: "default" | "small";
  children?: React.ReactNode;
  icon?: () => React.ReactNode;
  iconPosition?: "left" | "right";
  onClick?: (event: React.MouseEvent<HTMLElement>) => void;
  className?: string;
}
```

#### `Button` from CockroachCloud
```jsx
interface ButtonProps {
  category: ButtonCategory;
  actionType?: ButtonActionType;
  id?: string;
  className?: string;
  text?: string;
  href?: string; // external link
  to?: string; // internal link
  download?: string;
  small?: boolean;
  icon?: React.ReactNode;
  iconPosition?: ButtonIconPosition;
  disabled?: boolean;
  fluid?: boolean;
  onClick?: (...params: any) => any;
}
```

j-low:
# ui-components // button component

* resolves #21 
* implements button component, tests, and story

#### Checklist

### DIFF
--- a/packages/storybook-ui-components/stories/Button.stories.tsx
+++ b/packages/storybook-ui-components/stories/Button.stories.tsx
@@ -1,0 +1,85 @@
+import React, { FunctionComponent, ReactNode } from "react";
+import { withKnobs, text, select } from "@storybook/addon-knobs";
+import { Button, ButtonIntent, ButtonSize } from "@cockroachlabs/ui-components";
+
+export default {
+  title: "Button",
+  components: Button,
+  decorators: [withKnobs],
+};
+
+const ButtonDisplaySection: FunctionComponent<{
+  children: ReactNode;
+}> = ({ children }) => (
+  <section style={{ display: "flex", justifyContent: "flex-start" }}>
+    {children}
+  </section>
+);
+const ButtonDisplay: FunctionComponent<{ children: ReactNode }> = ({
+  children,
+}) => (
+  <div style={{ margin: "0rem 1rem", textAlign: "center" }}>{children}</div>
+);
+const Label: FunctionComponent<{ children: string }> = ({ children }) => (
+  <div style={{ marginBottom: "1rem" }}>
+    <code>{children}</code>
+  </div>
+);
+
+export const Example = () => (
+  <section>
+    <h1>Buttons</h1>
+    <section>
+      <h3>Type</h3>
+      <ButtonDisplaySection>
+        <ButtonDisplay>
+          <Label>primary-success</Label>
+          <Button intent="primary-success">Success</Button>
+        </ButtonDisplay>
+        <ButtonDisplay>
+          <Label>primary-danger</Label>
+          <Button intent="primary-danger">Danger</Button>
+        </ButtonDisplay>
+        <ButtonDisplay>
+          <Label>secondary (default)</Label>
+          <Button>Secondary</Button>
+        </ButtonDisplay>
+        <ButtonDisplay>
+          <Label>tertiary</Label>
+          <Button intent="tertiary">Tertiary</Button>
+        </ButtonDisplay>
+      </ButtonDisplaySection>
+    </section>
+    <section>
+      <h3>Size</h3>
+      <ButtonDisplaySection>
+        <ButtonDisplay>
+          <Label>standard (default)</Label>
+          <Button>Standard</Button>
+        </ButtonDisplay>
+        <ButtonDisplay>
+          <Label>small</Label>
+          <Button size="small">Small</Button>
+        </ButtonDisplay>
+      </ButtonDisplaySection>
+    </section>
+  </section>
+);
+
+export const Demo = () => {
+  const intents: Array<ButtonIntent> = [
+    "primary-success",
+    "primary-danger",
+    "secondary",
+    "tertiary",
+  ];
+  const sizes: Array<ButtonSize> = ["standard", "small"];
+  return (
+    <Button
+      intent={select("Intent", intents, "primary-success")}
+      size={select("Size", sizes, "standard")}
+    >
+      {text("Text", "button")}
+    </Button>
+  );
+};
--- a/packages/storybook-ui-components/yarn.lock
+++ b/packages/storybook-ui-components/yarn.lock
@@ -878,17 +878,19 @@
     lodash "^4.17.13"
     to-fast-properties "^2.0.0"
 
-"@cockroachlabs/icons@^0.2.1":
-  version "0.2.1"
-  resolved "https://registry.yarnpkg.com/@cockroachlabs/icons/-/icons-0.2.1.tgz#a8fcc05bbca1d434276a17be557b73e74c8d782f"
-  integrity sha512-hBhJ4343DBAxT5Zk2WmRdVJ+sfktZhrHNaTSFJs3XDDM2yXSPwdMphKN+NpPcH2z+SyyQZouzW3V3phMUNFMLw==
+"@cockroachlabs/icons@^0.2.3":
+  version "0.2.3"
+  resolved "https://registry.yarnpkg.com/@cockroachlabs/icons/-/icons-0.2.3.tgz#bd7f62a75e561f2f579bae413a757152a7f1bc24"
+  integrity sha512-/GuRDK9kwz/vbJyQBmM3xaxowl1De36pd6KAfobLGcMOZH4QEld0+nHRKyyrm5h0xm/VQaeMfG2veWHXfhooyw==
 
-"@cockroachlabs/ui-components@^0.1.4":
-  version "0.1.4"
-  resolved "https://registry.yarnpkg.com/@cockroachlabs/ui-components/-/ui-components-0.1.4.tgz#7ee8f7ae1731d29f35e9901375fa5c3f1570ab17"
-  integrity sha512-HKg1v2aTy3piLUrmBXJbiKXIUKtxEOKkAJcIL/0kTjKDSC2i0UzbxxAJHP8QZw09K4By5v/EK1F0G59vlwhZQg==
+"@cockroachlabs/ui-components@^0.1.6":
+  version "0.1.6"
+  resolved "https://registry.yarnpkg.com/@cockroachlabs/ui-components/-/ui-components-0.1.6.tgz#6b08e30c913a2893bf0cc86cda1a183370c72597"
+  integrity sha512-WpOtKf4y7TfmqbMLVyJTncxHURywxO0BdOMwZjml0LXnjbsdeKcEqyYsGGRnVYcPTiNDiDTKOs5o3aatZlpNWg==
   dependencies:
-    "@cockroachlabs/icons" "^0.2.1"
+    "@cockroachlabs/icons" "^0.2.3"
+    "@popperjs/core" "^2.4.0"
+    react-popper "^2.2.3"
 
 "@emotion/cache@^10.0.27", "@emotion/cache@^10.0.9":
   version "10.0.29"
@@ -1009,6 +1011,11 @@
   version "1.1.3"
   resolved "https://registry.yarnpkg.com/@nodelib/fs.stat/-/fs.stat-1.1.3.tgz#2b5a3ab3f918cca48a8c754c08168e3f03eba61b"
   integrity sha512-shAmDyaQC4H92APFoIaVDHCx5bStIocgvbwQyxPRrbUY20V1EYTbSDchWbuwlMG3V17cprZhA6+78JfB+3DTPw==
+
+"@popperjs/core@^2.4.0":
+  version "2.4.2"
+  resolved "https://registry.yarnpkg.com/@popperjs/core/-/core-2.4.2.tgz#7c6dc4ecef16149fd7a736710baa1b811017fdca"
+  integrity sha512-JlGTGRYHC2QK+DDbePyXdBdooxFq2+noLfWpRqJtkxcb/oYWzOF0kcbfvvbWrwevCC1l6hLUg1wHYT+ona5BWQ==
 
 "@reach/router@^1.2.1":
   version "1.3.3"
@@ -6651,6 +6658,14 @@ react-popper@^1.3.7:
     popper.js "^1.14.4"
     prop-types "^15.6.1"
     typed-styles "^0.0.7"
+    warning "^4.0.2"
+
+react-popper@^2.2.3:
+  version "2.2.3"
+  resolved "https://registry.yarnpkg.com/react-popper/-/react-popper-2.2.3.tgz#33d425fa6975d4bd54d9acd64897a89d904b9d97"
+  integrity sha512-mOEiMNT1249js0jJvkrOjyHsGvqcJd3aGW/agkiMoZk3bZ1fXN1wQszIQSjHIai48fE67+zwF8Cs+C4fWqlfjw==
+  dependencies:
+    react-fast-compare "^3.0.1"
     warning "^4.0.2"
 
 react-select@^3.0.8:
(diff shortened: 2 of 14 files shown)


packages/ui-components/src/Badge/Badge.module.scss
```diff
@@ -14,7 +14,7 @@
 
 .intent-neutral {
   background: $crl-neutral-2;
-  color: $crl-base-text--dark;
+  color: $crl-base-text-dark;
```
j-low, line 17: Did this because other token values are simply single-hyphen whereas the text values were the only token double-hyphened (in the BEM-flavored style).  I updated this (here and in tokens file) for consistency.

packages/ui-components/src/Button/Button.module.scss
```diff
@@ -0,0 +1,178 @@
...
+      background-color: $crl-base-green;
+      border-width: 1px;
+      border-color: mix(black, $crl-green-4, 20%);
+      @include crl-box-shadow(0, 0, 0, 3px, rgba($crl-base-green, 0.4));
+    }
+
+    &:disabled {
+      background-color: $crl-green-2;
+      border-color: $crl-green-2;
+    }
```
j-low, line 68: I wanted to DRY this code out as there is a lot of near-repetition in the related classes below but it turns out there actually isn't consistent relationships between the BG/border/color changes across element states across button types that would lend itself to abstracting the login to a handy mixin.  Alas.

packages/ui-components/src/styles/tokens.scss
```diff
@@ -122,16 +132,75 @@ $crl-base-divider: $crl-neutral-5;
...
+      for example in lightening a button in a disabled state.
+
+    Example usage:
+      ```
+      .my-css-class {
+        background-color: crl-lighten($crl-base-green);
+      }
+      ```
+  */
+  @return mix(white, $color, 70%);
```
j-low, line 185: Update to 90%.

Tried 70% in managed console and it's way too saturated still.  90% sits properly against the primary tone.

packages/storybook-ui-components/yarn.lock
```diff
@@ -878,6 +878,7 @@
     lodash "^4.17.13"
     to-fast-properties "^2.0.0"
 
+<<<<<<< HEAD
```
laurenbarker, line 881: merge conflict

packages/storybook-ui-components/stories/Button.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+export const Demo = () => (
+  <Button
+    intent={select(
+      "Intent",
+      ["primary-success", "primary-danger", "secondary", "tertiary"],
+      "primary-success",
+    )}
+    size={select("Size", ["standard", "small"], "standard")}
+    // disabled={select("Disabled", ["true", "false"], "false")}
```
laurenbarker, line 77: Delete?

packages/ui-components/src/Button/Button.module.scss
```diff
@@ -0,0 +1,221 @@
+@import "../styles/tokens.scss";
+
+.button {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  border-width: 1px;
+  border-radius: $crl-border-radius;
+  outline: none !important; // focused-state indicators added below
```
laurenbarker, line 9: Why is `!important` needed?

packages/ui-components/src/Button/README.md
```diff
@@ -0,0 +1,85 @@
...
+  <Button>Default button</Button>
+  <Button type="primary-success">Success button</Button>
+  <Button type="primary-danger">Danger button</Button>
+  <Button type="secondary">Secondary button</Button>
+  <Button type="tertiary">Tertiary button</Button>
+  <Buttont size="small">Small button</Button>
+);
+```
+
+## Properties
```
laurenbarker, line 21: Having this conversation on [another PR](https://github.com/cockroachdb/ui/pull/41#discussion_r434848133) with @nathanstilwell -- I don't think documenting all properties of a component in a README is sustainable. The docs addon should allow us to add this commentary in the [props table](https://github.com/storybookjs/storybook/blob/next/addons/docs/docs/props-tables.md).

packages/ui-components/src/README.md
```diff
@@ -2,3 +2,4 @@
 
 - [Badge](Badge/README.md)
```
laurenbarker, line 3: This README is already missing components (avatar and input). I don't think maintaining this list is sustainable either.

packages/ui-components/src/Button/__snapshots__/Button.test.tsx.snap
```diff
@@ -0,0 +1,12 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Button: render content can render using children 1`] = `
```
laurenbarker, line 3: Since we are planning enzyme I don't think we should add any snapshots. Shallow rendering won't be an option and we will just remove it anyway.

packages/ui-components/src/Button/README.md
j-low: Great.  I definitely agree.  Root-level READMEs are already vectors for stale information to creep in; collections of them will almost certainly get stale very, very rapidly without laser-focused CR on that particular aspect (which seems like a bad use of time).

packages/ui-components/src/Button/Button.module.scss
j-low: Leftover working state.  Good catch.

packages/ui-components/src/Button/Button.tsx
```diff
@@ -0,0 +1,48 @@
+import React, { FunctionComponent } from "react";
+import classnames from "classnames/bind";
+
+import objectToClassnames from "../utils/objectToClassnames";
+
+import styles from "./Button.module.scss";
+
+type ButtonIntent =
```
nathanstilwell, line 8: I think these should be exported to type the props in the Button Stories.

packages/storybook-ui-components/stories/Button.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+      </ButtonDisplaySection>
+    </section>
+  </section>
+);
+
+export const Demo = () => (
+  <Button
+    intent={select(
+      "Intent",
+      ["primary-success", "primary-danger", "secondary", "tertiary"],
```
nathanstilwell, line 73: This array could be typed (`Array<ButtonIntent>`) to prevent any typos.

packages/ui-components/src/Button/Button.tsx
```diff
@@ -0,0 +1,48 @@
...
+  | "primary-success"
+  | "primary-danger"
+  | "secondary"
+  | "tertiary";
+type ButtonType = "button" | "submit" | "reset";
+type ButtonSize = "standard" | "small";
+
+interface ButtonProps {
```
nathanstilwell, line 16: I had a conversation with Andrii about strictly typing the props of the Button component to allow for the _unhandled props_ to be used without specifying every possible prop of an HTML `<button>`. His point is that if we are using strict TypeScript we shouldn't allow untyped props to be passed to the the underlying button element. I see his point, and so the compromise we arrived at was to intersect the Component props with the underlying HTML element props. Something like,

```jsx
type OwnButtonProps = {
  intent?: ButtonIntent;
  type?: ButtonType;
  size?: ButtonSize;
};

type NativeButtonProps = Omit<
  React.ButtonHTMLAttributes<HTMLButtonElement>,
  keyof OwnButtonProps
>;

export type ButtonProps = NativeButtonProps & OwnButtonProps;
```

You will notice this also reduces the repetition of underlying props like `id`, `ariaLabel`, and `onClick`. Since we are also mirroring the `type` property of the underlying `ButtonHTMLAttributes` we could also remove our own `ButtonType` as well, thus boiling  the component down to the purely cosmetic expression that the component really is. Because we aren't really adding any functionality to an HTML button, we are just controlling what it's styles are.

packages/ui-components/src/Button/Button.tsx
```diff
@@ -0,0 +1,48 @@
...
+  type = "button",
+  size = "standard",
+  children,
+  ...props
+}) => {
+  const classNames = cx("button", objectToClassnames({ intent, size }));
+
+  if (children !== undefined) {
+    return (
+      <button className={classNames} type={type} {...props}>
```
nathanstilwell, line 39: Based on my comment above, here we would keep the _rest spread_ of props, but now those props are properly typed. And we could drop the `type` prop to simple be

```jsx
<button className={className} {...props}>
```

packages/ui-components/src/README.md
```diff
@@ -1,4 +1 @@
-# Components
-
-- [Badge](Badge/README.md)
-- [Tooltip](Tooltip/README.md)
+# ui-components
```
nathanstilwell, line 1: I guess I ultimately agree with @laurenbarker about the maintainability of the README documentation. Since we are going to move all documentation to Storybook, we might as well chop all README documentation out. I hope there is a way to generate something in the repo in terms of documentation, but until we get there this file doesn't make any sense anymore.

nathanstilwell: That being said, we should put A LOT more links to where Storybook is if it's going to be the only source of documentation for the components.

laurenbarker: +1 to adding a link to https://core-components.crdb.io/ in this readme

packages/ui-components/src/Button/Button.tsx
j-low: Great suggestion.  That solves that issue nicely without creating huge type definitions.

j-low: Note: I ended up putting the type prop back in.  The tests that set it via "setProp" fail but I don't want to hold things up so I've added a TODO to remove it and factor the tests when I can circle back around.

packages/ui-components/src/Button/Button.tsx
```diff
@@ -0,0 +1,53 @@
...
+  keyof OwnButtonProps
+>;
+
+export type ButtonProps = OwnButtonProps & NativeButtonProps;
+
+const cx = classnames.bind(styles);
+
+export const Button: FunctionComponent<ButtonProps> = ({
+  intent = "secondary",
+  type = "button",
```
nathanstilwell, line 35: Since `type` is intersected from the native props, we can default the value without having it be a part of `OwnButtonProps`. But you **wil** need to include it in the button element ....

packages/ui-components/src/Button/Button.tsx
```diff
@@ -0,0 +1,53 @@
...
+  size = "standard",
+  children,
+  ...props
+}) => {
+  const classNames = cx("button", objectToClassnames({ intent, size }));
+
+  if (children !== undefined) {
+    return (
+      <button className={classNames} {...props}>
```
nathanstilwell, line 44: here,

```suggestion
      <button className={classNames} type={type} {...props}>
```

packages/ui-components/src/Button/Button.tsx
```diff
@@ -0,0 +1,53 @@
+import React, { FunctionComponent, ButtonHTMLAttributes } from "react";
+import classnames from "classnames/bind";
+
+import objectToClassnames from "../utils/objectToClassnames";
+
+import styles from "./Button.module.scss";
+
+export type ButtonType = "button" | "submit" | "reset";
```
nathanstilwell, line 8: I think we can drop this because it already exist in `ButtonHTMLAttributes`.

```suggestion
```

packages/ui-components/src/Button/Button.tsx
```diff
@@ -0,0 +1,53 @@
...
+export type ButtonIntent =
+  | "primary-success"
+  | "primary-danger"
+  | "secondary"
+  | "tertiary";
+export type ButtonSize = "standard" | "small";
+
+type OwnButtonProps = {
+  // TODO(j-low): remove `type` prop and refactor tests accordingly
+  type?: ButtonType;
```
nathanstilwell, line 18: duplicated in `ButtonHTMLAttributes`

```suggestion
```

PR #56, opened 2020-05-18
Adjusting PR github action

nathanstilwell:
I'm renaming the action to be a little more descriptive and adding more
steps with names to make this action a little easier to see what's going
on.

I have added a step to build each package after bootstrap. This will
prevent errors where a package is linked, but the dist/ folder may not
exist and so the package will appear to not exist or not be installed
correctly.

### DIFF
--- a/.github/workflows/pull-request-ci.yml
+++ b/.github/workflows/pull-request-ci.yml
@@ -1,6 +1,6 @@
-name: Node.js CI
+name: Pull Request CI
 
-on: [push]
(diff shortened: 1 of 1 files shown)


laurenbarker:
Oh this is still a draft lol. Well renaming the action makes sense to me!

nathanstilwell:
Thanks for the review @laurenbarker I just put it in draft until I update some docs, but I think the workflow is finished.

PR #65, opened 2020-06-10
[icons] Simplying the build. Adding TypeScript.

nathanstilwell:
# Generated Icons

In this pull request I am simplifying the use of [svgr](https://react-svgr.com/) by removing Webpack from the equation and making this a two step process. Step one is generating react components from svg files. Step two is transpiling those components into JavaScript using the TypeScript compiler.

## Background

My first attempt at this package was a naive attempt to generate components from svg files. I wanted to get something done quickly and understood Webpack well enough at the time and so I put together something that worked. This solution had two issues. 

The first was that type checking would fail when referencing an svg file as a module. This was annoying but didn't really cause any critical errors as it would be compiled away when the files were bundled for publishing.

The second issue was that there were no types included in the published package and so consuming applications using TypeScript would balk that the components weren't typed. This would cause build failures in consuming applications.

## Removing Webpack

In my first attempts to fix these issues I created a [module declaration for the svgs](https://duncanleung.com/typescript-module-declearation-svg-img-assets/) but still had not type declarations to send with the publish bundle. Perhaps with my experience I may author the declaration files from scratch, but would rather the TypeScript compiler do it for me. Using a Webpack plugin to generate the React Components during a Webpack build didn't give the TypeScript compiler anything to generate declarations from. It was around this point that I realized Webapck was doing more harm than good.

With this clarity, I removed all the Webpack dependencies and installed [`@svgr/cli`](https://react-svgr.com/docs/cli/) and [TypeScript](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes.html#installing-typescript). Then using [`scripts`](https://docs.npmjs.com/cli-commands/run-script.html) I created a build that would create unbundled React Components along with type declarations along side of them.

```js
  "scripts": {
    "build": "npm-run-all build:generate build:typescript",
    "build:typescript": "tsc",
    "build:generate": "svgr --typescript --icon -d src/components svg",
    // ...
  },
```

The _generate_ step take any `*.svg` file in the `svg` folder and generates a React Component in the specified output folder (the `-d` option) `src/components`. The `--typescript` option adds types definitions and changes the output extension to `.tsx`. The `--icon` option sets the `height` and `width` to `1em` and preserves the viewbox.

The _typescript_ step runs the TypeScript compiler (`tsc`). All options are set in the `tsconfig.json`.

```json
{
  "compilerOptions": {
    "allowJs": false,
    "sourceMap": false,
    "noImplicitAny": true,
    "module": "commonjs",
    "moduleResolution": "Node",
    "target": "es6",
    "jsx": "react",
    "esModuleInterop": true,
    "declaration": true,
    "outDir": "./dist",
  },
  "exclude": [
    "dist",
  ],
}
```

Some notable options here are  `declaration`, `outDir`, and `module`. `declaration` is set to `true` to output `.d.ts` files along side the transpiled `.js` files. `outDir` is set to `dist` which is referenced by the package.json to build the published artifact. And we are setting `module` to `commonjs` because we intend this package to be installed via npm.

The resulting directory structure will look something like this after running `npm run build` (_I guess you could also use `yarn`_),

```text
├── AUTHORS
├── CONTRIBUTING.md
├── README.md
├── dist
│   ├── components
│   │   ├── ArrowLeft.d.ts
│   │   ├── ArrowLeft.js
│   │   ├── Aws.d.ts
│   │   ├── Aws.js
│   │   ├── Backup.d.ts
│   │   ├── Backup.js
│   │   ├── etc, etc, etc ...
│   │   ├── index.d.ts
│   │   └── index.js
│   ├── index.d.ts
│   └── index.js
├── notests.js
├── package.json
├── src
│   ├── components
│   │   ├── ArrowLeft.tsx
│   │   ├── Aws.tsx
│   │   ├── etc, etc, etc ...
│   │   └── index.tsx
│   └── index.ts
├── svg
│   ├── arrow-left.svg
│   ├── aws.svg
│   ├── etc, etc, etc ...
├── tsconfig.json
└── yarn.lock
```


packages/icons/CONTRIBUTING.md
```diff
@@ -0,0 +1,50 @@
+# Cockroach Icons
```
laurenbarker, line 1: I have never seen contributing files that aren't in the project repository's root, `docs`, or `.github` folder. I think all of this information should go in the readme. This is related to the conversation around what a contributing doc is for vs what a readme is for: https://github.com/cockroachdb/ui/issues/36#issuecomment-636073556

packages/icons/package.json
```diff
@@ -3,28 +3,30 @@
...
   "main": "dist/index.js",
+  "types": "dist/index.d.ts",
   "scripts": {
-    "build": "webpack",
-    "build:watch": "webpack --watch",
-    "clean": "rimraf dist",
+    "build": "npm-run-all build:generate build:typescript",
+    "build:typescript": "tsc",
+    "build:generate": "svgr --typescript --icon -d src/components svg",
+    "build:watch": "tsc --watch",
```
laurenbarker, line 14: Should the icons be regenerated on change? Not entirely sure what `tsc --watch` is doing

packages/icons/package.json
```diff
@@ -3,28 +3,30 @@
...
   ],
   "main": "dist/index.js",
+  "types": "dist/index.d.ts",
   "scripts": {
-    "build": "webpack",
-    "build:watch": "webpack --watch",
-    "clean": "rimraf dist",
+    "build": "npm-run-all build:generate build:typescript",
+    "build:typescript": "tsc",
+    "build:generate": "svgr --typescript --icon -d src/components svg",
```
laurenbarker, line 13: Wondering if it would be more clear if the generated components were added to `dist` instead of `src`. Is there a reason they are added to `src`?

PR #74, opened 2020-06-18
Prototype better Storybook story layout

nathanstilwell:


nathanstilwell:
Burning this down and trying again in #88 🔥 🔥 🧯

PR #89, opened 2020-07-08
Making some adjustments to the packages for less publish friction

nathanstilwell:
This PR is a result of some test publishing I did to try to understand the behavior of `lerna` when publishing. It should be noted that this PR involves version update to *live versions* (that already exist on npm).

**the backstory**

So the past few times I tried to publish in the UI repo the publish would fail. This was caused by having updated stories that referred to components that had not been published yet. It didn't make sense to me that Storybook was being build during publish as it is a private package. What I think was happening is during publish `lerna` looks at the dependency tree and marks all dependencies or dependents of packages that are being published. It then looks for `prepublish` scripts for the packages being published. It then seems to run the prepublish steps for all of the dependents and dependencies.

So in our case, in `ui-components` there is a `prepublishOnly` script that that runs `clean` and `build`. `ui-components` depend on `icons` and `design-tokens` and is a dependent of `storybook-ui-components`. So when there is a change that requires `ui-components` be published, `lerna` will run the prepublish steps on all of the dependents and dependencies before publishing. So this resulted in `clean` and `build` being run on all four packages when `ui-components` would be published. Sometimes it is worse, there was a `prepublish` step for `icons` also and so I noticed that `build` would be run twice on `icons` when publishing.

So I believe that Storybook is a special case. It is not a package that will ever be published and does not need to be _built_ during the npm lifecycle because it has not lifecycle. "Publishing" in the Storybook case is being deployed, which we have a github action for. So I have changed the name of the script from `build` to `build-storybook`, to denote this difference. The only time this script needs to get run is on a merge to master when the github action kicks off and so I don't think the inconsistency will effect either end users or ui developers.

~It dawns on me as I write all of this that we may still need a `prepublish` step for icons, as they technically need to be build before being published. The same is true for our design tokens. I suppose it's fine if they are build multiple times during a publish, but I may test this a little more before merging this PR.~ **Update** Everything works for now as expected. If we get some cowgirls publishing outside of lerna we can handle that on a case by case basis.

It also dawns on me I should have done all of this publishing as prerelease tags. Oh well.


packages/storybook-ui-components/package.json
```diff
@@ -1,11 +1,11 @@
 {
   "name": "storybook-ui-components",
-  "version": "0.2.0",
+  "version": "0.2.5",
   "description": "Storybook supporting ui-components as a component catalog",
   "main": "index.js",
   "private": true,
   "scripts": {
-    "build": "build-storybook -c .storybook -o public",
+    "build-storybook": "build-storybook -c .storybook -o public",
```
laurenbarker, line 8: Why was this changed? I thought we wanted all of our packages to use the same script names?

nathanstilwell: working on long description and back story in the PR description. Stay tuned.

laurenbarker: >changing `build` to be `build-storybook` to see if we can avoid building
storybook when publishing

Oh I see. I guess we don't really publish the storybook package

nathanstilwell: The TLDR in this case is, that storybook doesn't need to be built during development life cycle, only during deployment. So I think the inconsistency with other packages is appropriate here.

PR #91, opened 2020-07-13
ui-components: Publish to NPM registry with Github actions

koorosh:
Publishing to NPM registry is a multistep action with the following steps:
- run tests to ensure build is healthy;
- bump package with a new version, it creates a new commit in "master" branch and a tag for the new version
- push these changes (in "master" branch and tag)
- and finally publish to NPM

This action requires NPM Access Token to publish packages.
Token has to be set in secrets as YARN_TOKEN.

Note:
- extra commit is pushed to "master" branch in order to increment the version.
- commit is created under user name (who initiated this action) and with
empty ("<>") email because action context doesn't provide such information.

### DIFF
--- a/.github/workflows/publish-ui-components.yml
+++ b/.github/workflows/publish-ui-components.yml
@@ -1,0 +1,51 @@
+name: Publish ui-components to NPM Registry
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'packages/ui-components/**'
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
+      - name: Install dependencies
+        run: yarn install
+        working-directory: ./packages/ui-components
+      - name: Test
+        run: yarn ci
+        working-directory: ./packages/ui-components
+  publish:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-ui-components.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Install dependencies
+        run: yarn install
+        working-directory: ./packages/ui-components
+      - name: Configure Git User
+        run: |
+          git config user.name "${{ github.actor }}"
+          git config user.email "<>"
+      - name: Bump version (prerelease)
+        run: |
+          yarn version --prerelease
```
nathanstilwell, line 44: This was the part I haven't quite figured out. I would like to stick to _something_ like semantic versioning, but I'm not sure how a github action would be able to make that decision. In doing a prerelease version is it your intention that these are developer versions only? So incrementing the version would still happen manually?

koorosh: `yarn version --prerelease` command automatically:
- increments version in `package.json` for example `0.2.6-alpha.0` --> `v0.2.6-alpha.1`. `--prerelease` flag indicates which part to increment. Only major releases have to be done manually.
- Creates tag as the version number

So developers don't need to manually create versions.

PR #99, opened 2020-07-22
storybook-ui-components // Correcting fonts in stories

nathanstilwell:
Fonts were loading correctly, but were not being applied to stories via the
   "cockroach theme" (because that's not how the theme works). I think there
   had been an attempt to apply fonts to the stories by using a layout
   components, but had incorrectly used OpenSans instead of the correct Source
   Sans Pro.

   I am correcting the font and applying the layout component to all of the
   current stories. There's also a lot of "prettifying" that happened in the
   editor.

### DIFF
--- a/packages/storybook-ui-components/.storybook/preview-head.html
+++ b/packages/storybook-ui-components/.storybook/preview-head.html
@@ -1,4 +1,4 @@
 <link
-  href="https://fonts.googleapis.com/css?family=Roboto+Mono|Source+Sans+Pro&display=swap"
+  href="https://fonts.googleapis.com/css2?family=Roboto&family=Source+Sans+Pro:wght@400;600;700&display=swap"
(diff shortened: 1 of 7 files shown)


nathanstilwell:
Yes, I agree with this consensus (that we should be setting fonts via tokens). Was the consensus to set the font individually on every component? This feels odd to me as I would guess that most components should inherit font families from their parents. I think there are some font styles (`font-weight`, `font-size`, `font-style`, `line-height`) that make sense at a component level, but face seems strange to me to set on every component. As we tokenize fonts, I would love to discuss this further.

I also think we can push further the concept of _Layout components_ to take out the scattered inline CSS currently in our stories. This would play the role of the app setting the font face in an application. Just a thought.

rsadres:
This looks great. Thank you for helping me to get it running locally.

PR #105, opened 2020-07-29
ui-components // FuzzyTime Component

nathanstilwell:
# FuzzyTime

For the Notifications work being done by the Observability team, I needed to render _fuzzy times_ given timestamps (something like _"5 sec ago", "10 min ago"_,  etc). I felt like this is something we are likely to see again in designs and so I'm skipping ahead and adding it here.

Screen Shot 2020-07-29 at 14 34 41
Screen Shot 2020-07-29 at 14 35 14

#### Checklist
- [ ] ~I have updated the README of the package I'm working in to reflect my changes~ (I don't think we do this anymore


nathanstilwell:
I can't reproduce this build failure, so I'm going to try to PR again.

PR #106, opened 2020-07-29
ui-components // FuzzyTime Component

nathanstilwell:
# FuzzyTime

For the Notifications work being done by the Observability team, I needed to render _fuzzy times_ given timestamps (something like _"5 sec ago", "10 min ago"_,  etc). I felt like this is something we are likely to see again in designs and so I'm skipping ahead and adding it here.

Screen Shot 2020-07-29 at 14 34 41
Screen Shot 2020-07-29 at 14 35 14

#### Checklist
- [x] ~I have updated the README of the package I'm working in to reflect my changes~ (I don't think we do this anymore

### DIFF
--- a/packages/ui-components/src/FuzzyTime/FuzzyTime.tsx
+++ b/packages/ui-components/src/FuzzyTime/FuzzyTime.tsx
@@ -1,4 +1,4 @@
-import React, { FunctionComponent } from "React";
+import React, { FunctionComponent } from "react";
 import classnames from "classnames/bind";
 
 import { fuzzy } from "./util";
--- a/packages/ui-components/src/FuzzyTime/FuzzyTime.tsx
+++ b/packages/ui-components/src/FuzzyTime/FuzzyTime.tsx
@@ -5,16 +5,17 @@ import { fuzzy } from "./util";
 import styles from "./FuzzyTime.module.scss";
 
 export interface FuzzyTimeProps {
-  timestamp: string | Date;
+  timestamp: Date;
 }
 
 const cx = classnames.bind(styles);
 
 export const FuzzyTime: FunctionComponent<FuzzyTimeProps> = ({ timestamp }) => {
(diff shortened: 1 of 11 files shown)


nathanstilwell:
duplicate of #105. What a mess.

packages/ui-components/src/FuzzyTime/FuzzyTime.tsx
```diff
@@ -0,0 +1,20 @@
...
+import styles from "./FuzzyTime.module.scss";
+
+export interface FuzzyTimeProps {
+  timestamp: string | Date;
+}
+
+const cx = classnames.bind(styles);
+
+export const FuzzyTime: FunctionComponent<FuzzyTimeProps> = ({ timestamp }) => {
+  const timeago = fuzzy(new Date(timestamp));
```
koorosh, line 14: Should this component take care of parsing date from a string? Would it be better to expect the only Date and let app code to parse the string to date before passing it to the component?

nathanstilwell: The constructor for `Date` takes [either a string or a date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date) so I thought it be more user friendly to create a date from either.

I would love to hear about your thoughts here. Do you think this degrades type safety or creates confusion for the user rather than convenience?

koorosh: Accepting String type as input will require validation if an instantiated date is valid or not. In my opinion, it's not the responsibility of FuzzyTime component to parse/validate string to date format. 
With all of this, it brings extra complexity to decide what to show if the date is invalid.

nathanstilwell: I think that's a good point. If it somehow becomes too hard to pass a Date we can consider supporting a string, but no reason to do it until we have to.

PR #113, opened 2020-08-14
ui-components: Extending Avatar props

### LINKED ISSUE #112: Avatar improvements
In adding `Avatar` to Cockroach Cloud Console, a few improvement came up:

1) **Extend with div props to pass a `className` to extend the styling**

I needed to adjust the padding / margin of an `Avatar` instance and so I need to be able to add styles to the component. This is a pattern we already have on other ui-components, so it's just copying that pattern

2) **Avatar needs to take an `<svg>` as children**

Integrating into Cloud Console, and in some upcoming work, we need Avatar to be able to render an `<svg>` icon as content. I don't think there's any special consideration for positioning or rendering, but props will need to be adjusted from `children: "string"`

3) **Avatar needs to take an `` as children**

In discussing if Avatar should shorten a name to initials, we talked about the eventual use of an Avatar is to contain an image of a user. So Avatar should know how to handle an image tag as _children_.

nathanstilwell:
# ui-component // Extending Avatar props

I am extending the props from Avatar with attributes from HTMLDivElement.
This patten has been used in other components is necessary to extend
Avatar for integration with CockroachCloud console. I have also updated
the Avatar stories to demonstrate the use cases.

fixes #112

#### Checklist

### DIFF
--- a/packages/ui-components/src/Avatar/Avatar.tsx
+++ b/packages/ui-components/src/Avatar/Avatar.tsx
@@ -35,6 +35,7 @@ export const Avatar: React.FC<AvatarProps> = ({
   selectable = false,
   onClick,
   transformCase = "uppercase",
+  ...rest
 }) => {
   const classnames = useMemo(
     () =>
@@ -58,7 +59,7 @@ export const Avatar: React.FC<AvatarProps> = ({
   }, [onClick, disabled]);
 
   return (
-    <div className={classnames} onClick={onClickHandler}>
+    <div className={classnames} onClick={onClickHandler} {...rest}>
       {children}
     </div>
   );
--- a/packages/ui-components/src/Avatar/Avatar.tsx
+++ b/packages/ui-components/src/Avatar/Avatar.tsx
@@ -4,8 +4,7 @@ import classNames from "classnames/bind";
 import styles from "./Avatar.module.scss";
 import objectToClassnames from "../utils/objectToClassnames";
 
-export interface AvatarProps {
-  children?: string;
+interface OwnAvatarProps {
   size?: AvatarSize;
   intent?: AvatarIntent;
   disabled?: boolean;
@@ -13,6 +12,13 @@ export interface AvatarProps {
   onClick?: () => void;
   transformCase?: AvatarCase;
 }
+
+type NativeDivProps = Omit<
+  React.HTMLAttributes<HTMLDivElement>,
+  keyof OwnAvatarProps
+>;
+
+export type AvatarProps = NativeDivProps & OwnAvatarProps;
 
 export type AvatarSize = "default" | "small";
 export type AvatarIntent = "default" | "active" | "pending" | "invalid";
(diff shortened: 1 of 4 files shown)


packages/ui-components/src/Avatar/Avatar.tsx
```diff
@@ -14,6 +13,13 @@ export interface AvatarProps {
   transformCase?: AvatarCase;
 }
 
+type NativeDivProps = Omit<
```
laurenbarker, line 16: It doesn't look like any HTML attributes will be passed to the child div so I don't think this type is accurate.

nathanstilwell: Do you mean that since `OwnAvatarProps` doesn't overlap any of the `HTMLAttributes` (which is true) and that's why it's not accurate? Or do you mean that `React.HTMLAttributes<HTMLDivElement>` doesn't return any props? I'm adding this to use `className` and `children` from the underlying div prop types instead of defining them explicitly for Avatar.

laurenbarker: >I'm adding this to use className and children from the underlying div prop types instead of defining them explicitly for Avatar.

`children` is included with `React.FC` so this is just adding `className`.  

>Do you mean that since OwnAvatarProps doesn't overlap any of the HTMLAttributes (which is true) and that's why it's not accurate? 

Doesn't `onClick` overlap? I thought that was why `Omit` was being used. Why is Omit being used if there are no overlapping props?

`React.HTMLAttributes<HTMLDivElement>` includes a lot more than `className` which is why I don't think it's accurate. 

> Constructs a type by picking all properties from Type and then removing Keys. - [TS docs](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys)

```
Omit<
  React.HTMLAttributes<HTMLDivElement>, // all HTMLDivElement props
  keyof OwnAvatarProps // removing any overlapping keys
>;
```
Am I missing something?

nathanstilwell: Perhaps you are right. I was copying this pattern from other components because I thought this was a solved problem. Here's _what I want_,

1) `OwnAvatarProps` to be props we have defined to provide functionality of our component
2) `NativeDivProps` to be valid props of the underlying `<div>` element that will be rendered as the Avatar
3) `AvatarProps` to be the union type of both of these so that the user of `<Avatar>` can add any prop to the component either defined by our documentation or any valid prop of a `<div>`.

My expectation is that when using an `Avatar` that it should honor any prop that is valid html attribute for the underlying element. So that I can add a `className` or a data attribute or an `onClick`.

So I take it back, you are totally right, the `Omit` in this case is overriding the `onClick` from `React.HTMLAttributes<HTMLDivElement>` in favor our `onClick` prop.

But what is missing, that I think you are calling out, is that I want to pass along all of the rest of the props (besides className) that they user may have defined to the underlying `div`, like this

```jsx
const Avatar: React.FC<AvatarProps> = ({
  children,
  className,
  intent = "default",
  size = "default",
  disabled = false,
  selectable = false,
  onClick,
  transformCase = "uppercase",
  ...rest
}) => {
 {/*      snip          */}
   
 return (
    <div className={classnames} onClick={onClickHandler} {...rest}>
      {children}
    </div>
  );
};
```

PR #125, opened 2020-09-10
# Design Tokens // Adding types for design tokens

### LINKED ISSUE #109: [design-tokens] Add types for JS tokens
## What
Add types for JS tokens. See https://github.com/amzn/style-dictionary/issues/425#issuecomment-638010875 for an example of how we can generate type files.

## Why
Since we will be using the tokens in TS files, we need a type declaration file for the JS tokens.

nathanstilwell:
Generating a tokens.d.ts file for TypeScript support in applications. It
turns out that TypeScipt can infer types from exports that have
primative types, so we are just generating a duplicate of tokens.js with
a `.d.ts` extension and identifying it as `types` in package.json.

fixes #109

#### Checklist
~I have updated the README of the package I'm working in to reflect my changes~
~I have added or updated Storybook if appropriate for my changes~

### DIFF
--- a/packages/design-tokens/README.md
+++ b/packages/design-tokens/README.md
@@ -13,3 +13,43 @@ yarn
 ```
 npm run build
 ```
+
+## Exported Tokens
+
+### Sass
+
+`_tokens.scss` contains exported token values as Sass variables.
+
+```sass
+@import "~@cockroachlabs/design-tokens/dist/web/tokens"
+
+.example {
+    background-color: $color-background-button-primary-success-base;
+    color: $color-font-button-primary-success-base;
+    border-width: 1px;
+    border-style: solid;
+    border-color: $color-border-button-primary-success-base;
+}
+```
+
+### JavaScript
+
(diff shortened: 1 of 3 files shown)


packages/design-tokens/README.md
```diff
@@ -13,3 +13,43 @@ yarn
...
+// ...
+
+<SomeComponent
+  fontColor={ColorFont1}
+  interactionColor={ColorBaseBlue}
+  progressColor={ColorBasePurple}
+/>;
+```
+
+a `tokens.d.ts` file is also generate to act as types for tokens for TypeScript support.
```
j-low, line 55: nitty nit: "...is also generated..."

j-low (approved):
LGTM!  I love having tokens rolled up into typings.  Good stuff.

packages/design-tokens/README.md
nathanstilwell: cheers. 👍🏼

PR #129, opened 2020-09-17
ui-components: Button integration tweaks

### LINKED ISSUE #123: Design Review: Update button font
Update button component font to match Figma design system.

- It looks like the current implementation in story book uses Arial, and different size and letter spacing. 
Screen Shot 2020-08-23 at 1 33 50 PM

- Update to `font-family: Source Sans Pro`, `font-weight: 600 `, `letter-spacing: .02em`
Screen Shot 2020-08-23 at 1 37 12 PM

- This aligns with the `Body-strong` font in the Figma design system:

Screen Shot 2020-08-23 at 1 30 10 PM
Screen Shot 2020-08-23 at 1 29 48 PM

nathanstilwell:
# ui-component // Button

This PR is a collection of changes required to integrate the component into the CockroachCloud console application.

fixes #123 

#### Checklist

### DIFF
--- a/packages/ui-components/src/Button/Button.tsx
+++ b/packages/ui-components/src/Button/Button.tsx
@@ -7,7 +7,8 @@ import styles from "./Button.module.scss";
 
 export type ButtonIntent =
   | "primary"
-  | "primary-danger"
+  | "success"
+  | "danger"
   | "secondary"
   | "tertiary";
 
--- a/packages/ui-components/src/Button/Button.tsx
+++ b/packages/ui-components/src/Button/Button.tsx
@@ -27,9 +27,8 @@ type OwnButtonProps<T extends ElementType> = {
 // to be an Element. However, we also use `Omit` to remove any props that the element itself has if
 // we override them in `OwnButtonProps`. This lets us proxy an props that we'd like to manage
 // ourselves while keeping all the rest pure from the underlying element.
-type ButtonProps<T extends ElementType> =
-  OwnButtonProps<T> &
-  Omit<React.ComponentPropsWithoutRef<T>, keyof OwnButtonProps<T>>
+type ButtonProps<T extends ElementType> = OwnButtonProps<T> &
+  Omit<React.ComponentPropsWithoutRef<T>, keyof OwnButtonProps<T>>;
 
 const cx = classnames.bind(styles);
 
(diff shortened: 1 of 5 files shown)


packages/ui-components/src/Button/Button.tsx
```diff
@@ -1,50 +1,70 @@
...
   intent?: ButtonIntent;
   size?: ButtonSize;
-  onClick?: () => void;
+  fluid?: boolean;
 };
 
-type NativeButtonProps = Omit<
-  ButtonHTMLAttributes<HTMLButtonElement>,
-  keyof OwnButtonProps
+type ButtonProps<OwnProps, Type extends ElementType> = CombinedProps<
```
dhartunian, line 31: I think you can omit the double type parameter to `ButtonProps` and do:

```
type ButtonProps<Type extends ElementType> = CombinedProps<
  OwnButtonProps<Type>,
  ElementProps<Type>
>;
```

Because you want to make them always match.

In general, after playing around with this a bit, I don't think supporting an extra wide array of arbitrary HTML components is worth the type complexity.

I think we should just support `a` and `button` and extend from there. Then we can ensure a good experience with the component instead of making it infinitely customizable.

packages/ui-components/src/Button/Button.tsx
```diff
@@ -1,50 +1,70 @@
...
+  OwnProps,
+  ElementProps<Type>
```
dhartunian, line 33: I don't think you need `ElementProps` since its definition might be redundant. You can use `React.ComponentProps<Type>` directly since your definition of `ElementProps` picks all the keys from `React.ComponentProps`. If you want to omit some standard ones there appear to be helpers like `React.ComponentPropsWithoutRef<Type>`. Not sure what applies in this scenario.

dhartunian (commented):
Sorry this took so long, added some thoughts on the types.

packages/ui-components/src/Button/Button.tsx
nathanstilwell: I don't think we have that option (to only use `a` and `button`). We are internally linking with buttons all over the place.

dhartunian:
@nathanstilwell realizing after our chat that I should perhaps talk less and code more :) pushed up a change with my TS tweaks. Let's get this merged next sprint.

dhartunian:
Removed other reviewers for now, maybe there's more to clean up before circulating further. Leaving that up to you @nathanstilwell

PR #131, opened 2020-10-22
InlineAlert component

koorosh:
# ui-components // InlineAlert

InlineAlert component is one of basic components in Design System to
provide users with feedback on a page. It doesn't disappear and has
 to be placed somewhere on a page.

Screen Shot 2020-10-22 at 1 22 07 PM

#### Checklist

### DIFF
--- a/packages/ui-components/src/Icon/Icon.test.tsx
+++ b/packages/ui-components/src/Icon/Icon.test.tsx
@@ -2,7 +2,7 @@ import React from "react";
 import { shallow } from "enzyme";
 import { MinusCircle } from "@cockroachlabs/icons";
 
-import Icon, { IconProps } from "./Icon";
+import Icon, { IconProps, IconSize, IconTint } from "./Icon";
 
 const defaultProps = {
   iconName: "Plus",
@@ -10,7 +10,7 @@ const defaultProps = {
 
 const renderSubject = (overrideProps?: IconProps) => {
   const props = Object.assign(defaultProps, overrideProps);
-  return shallow(<Icon {...props} />);
+  return shallow<IconProps>(<Icon {...props} />);
 };
 
 test("Icon should render given an iconName", () => {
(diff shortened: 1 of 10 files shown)


packages/ui-components/src/Icon/Icon.test.tsx
```diff
@@ -41,8 +41,11 @@ describe("Icon size prop", () => {
     wrapper.setProps({ size: "small" });
     expect(wrapper.prop("className")).toContain("size-small");
 
-    wrapper.setProps({ size: "tiny" });
-    expect(wrapper.prop("className")).toContain("size-tiny");
+    wrapper.setProps({ size: "x-large" });
```
nathanstilwell, line 44: I appreciate the catch. I'm not sure why I added a tiny in here given that isn't a size we currently have. Would you mind re-ordering these from smallest to largest?

nathanstilwell (changes requested):
This looks great! I have one tiny change request, but everything else looks great. This is an exciting addition.

packages/ui-components/src/Icon/Icon.test.tsx
koorosh: What do you think about "dynamic" tests generation in this case? It might be easier to extend those test.
```typescript
const iconSizes = ["small", ... "x-large"]
iconSizes.forEach(iconSize => {
  test(`with ${iconSize} size`, () => { ... })
})
```

koorosh: @nathanstilwell , I've updated Icon tests (made more DRY)

packages/ui-components/src/Icon/Icon.test.tsx
```diff
@@ -31,18 +31,20 @@ describe("Icon size prop", () => {
...
-
-    wrapper.setProps({ size: "medium" });
-    expect(wrapper.prop("className")).toContain("size-medium");
-
-    wrapper.setProps({ size: "small" });
-    expect(wrapper.prop("className")).toContain("size-small");
-
-    wrapper.setProps({ size: "tiny" });
-    expect(wrapper.prop("className")).toContain("size-tiny");
+    const iconSizes: Array<[IconSize, string]> = [
```
nathanstilwell, line 34: I love this.

nathanstilwell (approved):
LGTM. Merging will publish a prerelease tag. I will manually publish a latest version.

PR #132, opened 2020-11-02
ui-components: Spinner component

koorosh:
# ui-components // Spinner

Spinner component is animated loading indicator which can be
used in tables or on pages and indicate intermediate state.

This component doesn't reuse existing Spinner icon due to
technical limitations. Loading icon is rendered with `<path>`
element which is difficult to animate and make it rotate around
SVG central point. That's why current component implements its
own loading icon based on `<circle>` element.

#### Checklist

output

For comparison, loading indicator which is used in Admin UI right now:
spinner


nathanstilwell (approved):
This looks great. Thank you @koorosh

PR #137, opened 2020-11-12
pkg design-tokens: Update design tokens as per rebrand

### LINKED ISSUE #134: [design-tokens] update colors for redesign
# Update Color tokens

Our basic color values have been updated for the rebranding effort (NOV 2020). Update [`core.json`](https://github.com/cockroachdb/ui/blob/master/packages/design-tokens/tokens/color/core.json) to match _Color_ section in Figma. These core tokens flow down to other tokens ([`base.json`](https://github.com/cockroachdb/ui/blob/master/packages/design-tokens/tokens/color/base.json), [`background.json`](https://github.com/cockroachdb/ui/blob/master/packages/design-tokens/tokens/color/background.json)), so you may need to spot check them to see if the results are expected. You generate a the sass variables by running `npm run build` in the `design-tokens` package.

poojamaniar:
# design-tokens // update as per rebrand

Changes include: 
- blue and purple color palettes have changed
- button primary colors have changed from green to purple

fixes #134

### DIFF
--- a/packages/design-tokens/tokens/color/background.json
+++ b/packages/design-tokens/tokens/color/background.json
@@ -2,8 +2,6 @@
   "color": {
     "background": {
       "base": { "value": "{color.base.white.value}" },
-      "alt": { "value": "{color.core.neutral.1.value}" },
-      "inverse": { "value": "{color.core.neutral.6.value}" },
       "success": { "value": "{color.base.green.value}" },
       "error": { "value": "{color.base.red.value}" },
       "warning": { "value": "{color.base.orange.value}" },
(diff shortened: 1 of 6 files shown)


poojamaniar:
There are a few things I'm unclear about. Open questions include:

1. Where do font colours get applied in console? Where were the values in ui-components generated from? 
2. A lot of border-colors, and some background-colors for buttons use mix() in console, and I'm wondering if this can be removed and a colour from our palette can be chosen instead. If I create a ticket, can we ask design to make this change?

packages/design-tokens/tokens/color/background.json
```diff
@@ -14,42 +12,32 @@
         "neutral": { "value": "{color.core.neutral.2.value}" },
         "success": { "value": "{color.core.green.1.value}" },
         "warning": { "value": "{color.core.orange.1.value}" },
-        "info": { "value": "{color.core.blue.1.value}" },
         "danger": { "value": "{color.core.red.1.value}" }
       },
 
       "avatar": {
-        "default": { "value": "{color.core.neutral.2.value}" },
+        "default, disabled": { "value": "{color.core.neutral.2.value}" },
```
nathanstilwell, line 19: this file needs to contain valid JSON, so comma separated keys aren't going to work. This would instead need to be, 
```
 "default": { "value": "{color.core.neutral.2.value}" },
 "disabled": { "value": "{color.core.neutral.2.value}" },
```

packages/design-tokens/tokens/color/border.json
```diff
@@ -15,30 +15,20 @@
...
         "primary": {
           "success": {
-            "base": { "value": "{color.base.green.value}" },
-            "hover": { "value": "{color.base.green.value}" },
-            "active": { "value": "{color.core.green.4.value}" },
-            "focused": { "value": "{color.core.green.4.value}" },
-            "disabled": { "value": "{color.base.green.value}" }
+            "base, hover": { "value": "{color.base.purple.value}" },
+            "hover, active, focused": { "value": "{color.core.purple.4.value}" },
+            "disabled": { "value": "{color.base.purple.2.value}" }
```
nathanstilwell, line 20: `color.base.purple.2` doesn't exist, perhaps you meant `color.core.purple.2.value`?

packages/ui-components/src/Button/Button.tsx
```diff
@@ -8,8 +8,7 @@ import styles from "./Button.module.scss";
 export type ButtonIntent =
   | "primary-success"
   | "primary-danger"
-  | "secondary"
-  | "tertiary";
```
nathanstilwell, line 12: What happened to tertiary intent? I still see it in Figma.

packages/design-tokens/tokens/color/background.json
poojamaniar: I knew it seemed too good to be true... updated.

packages/ui-components/src/Button/Button.tsx
poojamaniar: I see, in console it's called a flat button so I can keep the tertiary in ui-components and later update in console (probably when the button component is edited to be able to be used in console).

nathanstilwell (approved):
Looks Great!!

PR #141, opened 2020-11-20
Text component

nathanstilwell:
# Type Components

## Overview

Our design specifications outline a fixed number of font treatments for use across all applications and content. I am offering three type components based on the [`Text` component found in `admin-ui-components`](https://github.com/cockroachdb/admin-ui-components/blob/master/src/text/text.tsx); `<Heading>`, `<Text>`, `<Code>`.

## Design Specifications

There are a limited set of type treatments defined in our designs. I understand them to fall into three categories. 

### Heading

Screen Shot 2020-12-04 at 17 19 27

### Text
Screen Shot 2020-12-04 at 17 20 10

### Code
Screen Shot 2020-12-04 at 17 21 48

What makes this pattern especially attractive is that these styles are kept consistent across our designs in Figma using [Styles](https://help.figma.com/hc/en-us/articles/360039238753-Styles-in-Figma). For example where we see this heading style being use in this UI mock,

Screen Shot 2020-12-07 at 11 59 12

we can see on the left that the style **Heading/H5** is being applied. Which is consistent with the definitions found in the Design System specification.

Screen Shot 2020-12-07 at 12 09 49

This means that UI engineers can make a direct correlation between the Figma styles they see in mocks and the Typography components found in the `ui-components` package. In this case using `<Heading type="h5>Add Email Recipients</Heading>`

## Implementation

I started by breaking the values defined in these treatments into tokens. These fell into the categories of font family, size (of fonts and line-heights), and font weights. I defined these in token files in a `type` directory in that package. At first I was concerned about constraining certain weight to certain font faces, but realized that these constraints would be created by the components themselves. In any case browsers are able to gracefully fall-back when a weight isn't specified by a particular font-face.

The three components themselves are quite simple, each taking a single prop to define their treatment. In the case of `<Heading>` the `type` matches to the rendered HTML element (`h1`...`h6`) as these elements are quite meaningful in terms of information architecture and accessibility. For `<Text>`, the types are defined by their intended usages as specified by design (`body`, `caption`, etc). The `<Text>` component will render a `span` as this is the most generic inline element and can safely be the child of any other HTML element. In the case of `<Code>`, the treatments are identical aside from a font weight, and so I thought a prop of `weight` would make the most sense. `<Code>` will render a `<code>` element and so maybe be subject to global styles. But that is up to the implementing application to determine if this is an advantage or a hindrance.

#### Checklist
- [x] ~I have updated the README of the package I'm working in to reflect my changes~

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/ui/141)


packages/ui-components/src/Type/Text.tsx
```diff
@@ -0,0 +1,41 @@
...
+export const Text = ({
+  type = "body",
+  noWrap = false,
+  className,
+  children,
+  ...rest
+}: TextProps) => {
+  const classNames = cx(
+    objectToClassNames({ type }),
+    { ".no-wrap": noWrap },
```
koorosh, line 32: ".no-wrap" class name should be provided here without dot, it's not recognized properly.
Would you mind to extend Demo storybook with `nowrap` option?

packages/ui-components/src/index.ts
```diff
@@ -7,3 +7,4 @@ export * from "./Icon/Icon";
 export * from "./FuzzyTime/FuzzyTime";
 export * from "./InlineAlert/InlineAlert";
 export * from "./Spinner/Spinner";
+export * from "./Type/";
```
koorosh, line 10: It took me a while to find this components under `Type` dir 🙂 . What do you think about renaming it to `Typography` or something more specific, because it sounds too generic.

packages/design-tokens/tokens/type/family.json
```diff
@@ -0,0 +1,12 @@
+{
+  "type": {
+    "family": {
+      "ui": {
+        "value": "\"Source Sans Pro\", sans-serif"
```
koorosh, line 5: @nathanstilwell should `design-tokens` package include font files as well or client app is responsible for providing them?

packages/ui-components/src/index.ts
nathanstilwell: I like that suggestion. "Type" is a pretty overloaded term in a _TypeScript_ codebase. 😄

packages/design-tokens/tokens/type/family.json
nathanstilwell: Since we use Google Fonts to handle our licensing and load our font files, I believe it is overly complicated to have components ship with their own font files or include them with design tokens. For now I would like to keep the burden of loading fonts on the implementing application.

packages/ui-components/src/Type/Text.tsx
nathanstilwell: Good catch on both counts. I thought the `noWrap` prop was a nice convenience but then forgot to document it in Storybook.

nkodali (commented):
> Reviewed 3 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/ui/141)* status: 3 of 13 files reviewed, 4 unresolved discussions (waiting on @dhartunian, @j-low, @koorosh, @laurenbarker, @nathanstilwell, and @nkodali)

___
*[packages/ui-components/src/Type/Code.tsx, line 1 at r1](https://reviewable.io/reviews/cockroachdb/ui/141#-MNuzYhi4BcivAmkW2mp:-MNuzYhi4BcivAmkW2mq:b-k9ca73) ([raw file](https://github.com/cockroachdb/ui/blob/138c9bc1528da3700826d8501cee1f3892ed60e5/packages/ui-components/src/Type/Code.tsx#L1)):*
> ```XML
> import React from "react";
> ```

It seems like we're using the convention to have capitalized directory names with same name as the component within the directory. Should this convention apply when there are multiple different components in the directory as well?

nathanstilwell (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/ui/141)* status: 2 of 14 files reviewed, 4 unresolved discussions (waiting on @dhartunian, @j-low, @koorosh, @laurenbarker, and @nkodali)

___
*[packages/ui-components/src/Type/Code.tsx, line 1 at r1](https://reviewable.io/reviews/cockroachdb/ui/141#-MNuzYhi4BcivAmkW2mp:-MNy_Tmw7mmPFEjI1Ii2:b-3reasn) ([raw file](https://github.com/cockroachdb/ui/blob/138c9bc1528da3700826d8501cee1f3892ed60e5/packages/ui-components/src/Type/Code.tsx#L1)):*
<details><summary><i>Previously, nkodali wrote…</i></summary><blockquote>

It seems like we're using the convention to have capitalized directory names with same name as the component within the directory. Should this convention apply when there are multiple different components in the directory as well?
</blockquote></details>

I'm not sure I've landed on a hard and fast rule when grouping components into a directory. We did this with `Input` where Andrii has a couple of component living in a single directory. Are you referring to the capitalization or the directory? I am of the opinion that the directory and file name should be capitalized since the primary export is a React component and must be capitalized. And I typically try to match the filename to the default export. I suppose I extended this rule that a directory containing primarily React components could be capitalized as well. 

I am open a _bikeshed_ conversation to land on the best convention for this kind of thing.

Annebirzin:
@nathanstilwell Looks great! A couple of minor things I noticed:
- H6 should be `line-height: 16 `
- Caption should be `line-height: 20`
- Caption bold should be `line-height: 20`
- The  labels for the code section should be updated to reflect `Code regular` `Code medium` and `Code bold`

nathanstilwell:
>     * The  labels for the code section should be updated to reflect `Code regular` `Code medium` and `Code bold`

👀 👀 👀 

Screen Shot 2020-12-07 at 16 45 45

🤦🏼  Looks like someone got a little too excited with the _copy pasta_. ha ha 🍝

koorosh (approved):
> Reviewed 2 of 13 files at r1, 7 of 10 files at r3, 3 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/ui/141)* status: all files reviewed, 1 unresolved discussion (waiting on @dhartunian, @j-low, @laurenbarker, and @nkodali)

PR #142, opened 2020-11-24
[DS-13] ui: Form fields and inputs (refactor / add)

poojamaniar:
# ui-components // Form inputs and fields
----------------------------------------------------------------------------------------------------------------
Changes include: 
- modifying the BaseInput component to be the common error and help handling, common to all inputs (except Select)
- one base Field component which can render types of Inputs (Text, Checkbox, Password) -> this is used in each custom TextField, EmailField, etc, to ensure that each input type's props are followed strictly. 
- refactor Text, add Checkbox, Email, Password inputs based on what's in console
- use react-final-form state instead of local state
- create New,Existing Password input types (inherits from Text, but with validation function passed in from console)
- SelectInput was copied over from console, no changes made, except where an Icon was rendered in the component, now it's passed in as a prop -> this is because the ui repo needs icons to be updated, also needs Icon component to be updated. 
- css changes for components
- refactored Icon component in ui-compoennts to be used in NewPasswordInput!! Yay.

Still to do:
- add react-final-form as a dependency of ui-components
----------------------------------------------------------------------------------------------------------------
Screenshots of Components:

ExistingPasswordField, NewPasswordField
image
EmailField
image
TextField
image
CheckboxField
image
----------------------------------------------------------------------------------------------------------------

Visual Architecture Changes:
image

Notable things from the before architecture:
- Some inputs were created as fields containing inputs (SelectInput, Input)
- Code repetition, InputField, CheckboxField, as well as Text and Checkbox
- SelectInput and Input didn’t use react final form to manage state (SelectInput)

Highlights from the architecture after are:
- CommonInput component was created, to define how error and help messages should be shown in all the components.
- More code being shared between components, through CommonInput, and Field components which are used by 4 different types of inputs
- All refactored to use react-final-form

----------------------------------------------------------------------------------------------------------------

#### Checklist


packages/ui-components/src/Input/EmailPasswordInput.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export const NewPasswordInput: React.FC<NewPasswordProps> = ({
+  ...props
+}) => {
+  return (
+    <BaseTextInput {...props}/>
+  );
+};
+
+export const ExistingPasswordInput: React.FC<ExistingPasswordProps> = ({
```
nathanstilwell, line 25: I find it confusing to have two separate components for a _NewPassword_ and an _ExistingPassword_. If the only real difference is the _forgot password_ link, then I think that could be a prop. Going a step further, why do we even need a boolean? If the user supplies a _forgot password_ element, then it should be displayed otherwise not. Then we could simply have a `<PasswordInput>` component.

packages/ui-components/src/Input/Field.tsx
```diff
@@ -0,0 +1,115 @@
...
+          {...inputProps}
+          textType={type}
+          error={meta.touched && meta.error}
+      />
+      }}
+    />
+  );
+};
+
+// It's not necessary to pass the type prop for a Field
```
nathanstilwell, line 92: I think it is a sensible requirement to require the user to define a "type" of Field. We could default it to "text". I would expect there to be a type defined for "FieldType" or something. Maybe something like,

```ts
type FieldType = "text" | "checkbox" | "email" | "password";
```

Then based on the value of this prop we could constrain the props and render the appropriate field.

packages/ui-components/src/Input/SelectInput.tsx
```diff
@@ -0,0 +1,312 @@
+import React, {
```
nathanstilwell, line 1: When Joe first introduced the `SelectInput` component, we considered adding it to ui-components then. In the end we decided that this component is too specialized to what it's doing and is unlikely to be reused outside of Cockroach Cloud Console without significant refactor to make it more generic.

Therefore, I don't believe we should be adding it to our common forms here.

packages/ui-components/src/Input/TextTypeInput.tsx
```diff
@@ -0,0 +1,170 @@
...
+  ariaLabelledBy?: string;
+  autoFocus?: boolean;
+  tabIndex?: number;
+  onKeyDown?: (
+    event: React.KeyboardEvent<HTMLInputElement>,
+  ) => void;
+  onClick?: (
+    event: React.MouseEvent<HTMLInputElement, MouseEvent>,
+  ) => void;
+  forgotPasswordLinkDiv?: JSX.Element;
```
nathanstilwell, line 54: I don't think we should limit this element to a "div". Since it's type is a JSX element, perhaps we should call it `forgotPasswordLinkElement`?

packages/ui-components/src/utils/objectToClassnames.ts
```diff
@@ -1,3 +1,27 @@
...
+import uniq from "lodash/fp/uniq";
+import isEmpty from "lodash/isEmpty";
+import negate from "lodash/negate";
+
+const appendSuffix = (suffix: string) => (string: string) =>
+  `${string}${suffix}`;
+
+const containerSuffix = appendSuffix("__container");
+
+export const generateContainerClassnames = (className = ""): string =>
```
nathanstilwell, line 16: I'm hoping that we don't need this utility. But if we do, it should be moved to it's own file of the same name.

nathanstilwell (changes requested):
Overall there seems to be quite a lot of composition going on, which in general isn't a bad thing. But I wonder if we might simplify this PR a bit by removing `SelectInput` (as I don't think it belongs in ui-components anyway) and separating `Checkbox` into a different component. I believe we could simplify the code a bit by exposing a `InputField` and a `CheckboxField` component.

Also, from the beautiful graph you provided in the description I wonder where is `BaseInput` in the graph? And how does it differ from `CommonInput`?

packages/ui-components/src/Input/Field.tsx
```diff
@@ -0,0 +1,115 @@
+import React, { FunctionComponent } from "react";
+import { Field, FieldProps, FieldRenderProps } from "react-final-form";
```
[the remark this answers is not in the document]
poojamaniar: Yup, wanted to separate them into a different PR incase it needed more discussion as to what dependency we want to add final form as (I put it in dev), linked here: https://github.com/cockroachdb/ui/pull/144

packages/ui-components/src/Input/BaseInput.tsx
```diff
@@ -1,102 +1,58 @@
...
-      value={value}
-      placeholder={placeholder}
-      className={classnames}
-      style={style}
-      onChange={onChangeHandler}
-      autoComplete={autoComplete}
-      disabled={disabled}
-    />
+    <div
+      className={classNames("crl-input__container", classes, {
```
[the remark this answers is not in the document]
poojamaniar: yea... i tried using just inline, or scrapping the container wrapper but I can't seem to get inline fields to render with 100% of the width. If I find a fix, will push.

packages/ui-components/src/Input/BaseInput.tsx
```diff
@@ -1,102 +1,58 @@
...
-  ChangeEvent,
-  CSSProperties,
-  useCallback,
-  useEffect,
-  useMemo,
-  useState,
-} from "react";
+import React from "react";
 import classNames from "classnames/bind";
-import styles from "./styles.module.scss";
```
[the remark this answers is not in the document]
poojamaniar: removed!

packages/ui-components/src/Input/EmailPasswordInput.tsx
poojamaniar: I remember discussing this in our meeting yesterday, but just to ensure that it's noted here for documentation:
The differences between Existing/New Password are:
1. New password can render as a text/password type, depending on if the icon is set to show/hide the password
2. suffix icon is provided in new password, to toggle the type
3. forgot password link is in existing password component
4. new password has text underneath the field to specify whether the password passes validation or not (EmailPasswordInput.tsx has the text and the styling associated with this).

poojamaniar: These differences weren't in the original few commits, but were realized after I tried to integrate the components into console.

packages/ui-components/src/Input/Field.tsx
poojamaniar: We has discussed a better way of returning the correct inputs (with their specified props) in the Field component, but I haven't found a better way yet. I recall Button may have used some of this logic - still have to look into that.

poojamaniar:
Has been moved from Draft status as the work currently being done is:
1. Removing BEM from styling.
2. Looking into Button component, or other examples of how to render specific Inputs while only accepting their prop types.

CommonInput has been renamed from BaseInput to indicate it renders the common things shared between components (error and help messages).

packages/ui-components/src/utils/objectToClassnames.ts
poojamaniar: removed, I think we can avoid it if we just change the classNames we pass in from console, which can be done in the integration. from what i see in console, it's only used in one form (deleteOrgModal)

packages/ui-components/src/Input/helpers.tsx
```diff
@@ -1,12 +1,12 @@
 import React from "react";
 import classNames from "classnames/bind";
 import styles from "./styles.module.scss";
-import { InputProps } from "./BaseInput";
+import { CommonInputProps } from "./CommonInput";
```
poojamaniar, line 4: This file is only used for NumberInput, referring to the NumberInput that was already in ui-components when I started this proj. However, I created another NumberInput by changing the type of TextTypeInput, and the prop type (from string to number) -- the new one is a simpler component, easier to understand. I've tested it a little, seems to work as expected.

 I've kept the old NumberInput (and thus this helpers file) incase the new one doesn't work for the use cases it's needed for.

packages/ui-components/src/Input/BaseInput.tsx
```diff
@@ -1,102 +1,58 @@
...
-  const onChangeHandler = useCallback(
-    (event: ChangeEvent<HTMLInputElement>) => {
-      const nextValue = event.target.value;
-      if (onChange && !disabled) {
-        onChange(nextValue);
-      }
-    },
-    [onChange, disabled],
+  const helpMsg = !help ? null : (
+    <div className="crl-input__message--info">{help}</div>
```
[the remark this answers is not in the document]
poojamaniar: BEM removed.

poojamaniar:
Build is failing because this requires #143 to be merged first (specifically for the icon intent prop to be passed).

packages/ui-components/package.json
```diff
@@ -28,6 +28,7 @@
   "dependencies": {
     "@cockroachlabs/icons": "^0.3.0",
     "@popperjs/core": "^2.4.3",
+    "final-form": "^4.20.1",
```
nathanstilwell, line 31: Instead of bundling `final-form` with the lib, I think we should declare it as a `peerDependency`.

poojamaniar: done

packages/ui-components/src/Input/CommonInput.tsx
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+import classNames from "classnames/bind";
+import "./input.module.scss";
+
+export interface CommonInputProps {
+  id?: string;
+  // this should be the div containing the input
+  fieldInput?: JSX.Element;
+  classes?: string;
+  className?: string;
```
nathanstilwell, line 10: Why do we have both `classes` and `className` here? I feel like `classes` is an unconventional choice. Is this a hold over from the CC console component?

packages/ui-components/src/Input/CheckboxInput.tsx
```diff
@@ -0,0 +1,55 @@
+import React from "react";
+import classNames from "classnames/bind";
+import { CommonInputProps, CommonInput } from "./CommonInput";
+
+interface CheckboxProps {
+  value?: boolean;
```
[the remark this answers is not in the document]
poojamaniar: alright - changed the input to expect a checked field (as the native input props have, instead of the boolean value type.

package.json
```diff
@@ -2,6 +2,7 @@
   "name": "root",
   "private": true,
   "devDependencies": {
+    "@cockroachlabs/ui-components": "^0.2.12",
```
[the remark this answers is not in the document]
poojamaniar: moved to peer dependencies, since we import from cockroachlabs/icons in Icon component (living inside ui-components).

packages/ui-components/src/Input/CommonInput.tsx
poojamaniar: className can be removed, it's not used

packages/ui-components/src/Input/CheckboxInput.tsx
```diff
@@ -0,0 +1,55 @@
+import React from "react";
+import classNames from "classnames/bind";
+import { CommonInputProps, CommonInput } from "./CommonInput";
+
+interface CheckboxProps {
```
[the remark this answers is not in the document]
poojamaniar: Okay - updated. This also caused there to be a new input type, multiline, which differs from single line text inputs because multiline renders a <textarea />, and thus accepts textarea props, which differ from <input /> props single line text inputs render.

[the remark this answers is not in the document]
poojamaniar: CommonInputProps will likely not need to extend input props, because it accepts an input component (such as text, or password) as a prop. the props CommonInput uses are only for styling the help / error messages, which is an internal convention, not related to native inputs, I believe...

Text, Checkbox have been updated to extend the native <input /> props.

packages/ui-components/src/Input/Field.tsx
nathanstilwell: I'm punting on this idea for now.

nathanstilwell (commented):
I'm having a hard time with this review. The lack of documentation makes it hard to verify the behavior of these components is correct. A process that I assume is happening by testing these components in an application. For other components in this library, we have had designs (in Figma) to compare functionality against, and that is only partially true in this case. Specifically there are no designs for a multi-line input and the concept of a "Field Group" in the designs isn't strongly represented in these components.

Overall, the code I feel is correct but I would like to see stories in Storybook for these components or some other kind of documentation about their intended use as a follow up PR.

packages/ui-components/src/Input/CommonInput.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+export interface CommonInputProps {
+  id?: string;
+  // this should be the element containing the input
+  fieldInput?: JSX.Element;
+  className?: string;
+  disabled?: boolean;
+  help?: string | JSX.Element;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  error?: any;
```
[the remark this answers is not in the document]
poojamaniar: This is set to any because error in an Input is assigned a value of` meta.touched && meta.error`, and `meta.error` is assigned a type of `any`. I'll add that comment in the code.

packages/ui-components/src/Input/CommonInput.tsx
```diff
@@ -0,0 +1,50 @@
...
+  inline?: boolean;
```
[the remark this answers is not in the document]
poojamaniar: I think it could, but it's most commonly used in SingleLineTextInputs, I haven't seen it used anywhere else. I couldn't find documentation about which Inputs it applies to so I assumed all -- perhaps not a correct assumption?

packages/ui-components/src/Input/EmailPasswordInput.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const [error, setError] = useState(meta && meta.error);
+  const [touched, setTouched] = useState(meta && meta.touched);
+
+  useEffect(() => {
+    setError(meta && meta.error);
+    setTouched(meta && meta.touched);
+  }, [meta]);
+
+  // validators is a wrapper array of validation functions
+  const validator = validators && validators.length > 0 && validators[0];
```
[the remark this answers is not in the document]
poojamaniar: we currently only use 0/1 validators, so instead of passing an array (which was only being used to get the validator label), i've just changed the props to conditionally accept the validator label.

packages/ui-components/src/Input/TextTypeInput.tsx
```diff
@@ -0,0 +1,176 @@
...
+  // this prop is only used for internal implementation
+  // whether the password exists or not, should be indicated by New/Existing PasswordInput
+  existingPassword?: boolean;
+  //The actual type below as a parameter is called FieldValue
+  // not exported by react-final-form, but by google protobuf
+  // which would need to be added as a dependency
+  // since this prop comes from the Field component of react-final-form
+  // we can be fairly certain that the type won't be unexpected
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  meta?: FieldMetaState<any>;
```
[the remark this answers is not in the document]
poojamaniar: I had typed it as `any` because I wasn't sure what `FieldValue` was in `FieldRenderProps` (which passed that value to `FieldMetaState`) - I looked into it and it's defined by the HTML Input component we pass in. Updated the code to reflect this.

nathanstilwell (approved):
I'm ready to merge this. I think there are some things we have the opportunity to iterate on, but there is a lot here that is valuable. Excellent work @poojamaniar. Thank you for being so diligent on this PR.

PR #143, opened 2020-12-04
ui: refactor icon component to use in Input components

poojamaniar:
# ui-components // refactor icon component

Tint is not used, replaced with intent, inline styling added for container incase container styling is needed (if Icon is more widely used in console, it will need container styling. Currently for being used in ui-component, in NewPasswordInput, we don't need container styling).

#### Checklist

### DIFF
--- a/packages/ui-components/src/Icon/Icon.module.scss
+++ b/packages/ui-components/src/Icon/Icon.module.scss
@@ -63,10 +63,10 @@
 }
 
 .fill-disabled {
-  fill: $color-code-neutral-5;
+  fill: $color-core-neutral-5;
 
   &-light {
-    fill: $color-code-neutral-4;
+    fill: $color-core-neutral-4;
   }
 }
 
--- a/packages/ui-components/src/Icon/Icon.module.scss
+++ b/packages/ui-components/src/Icon/Icon.module.scss
@@ -34,29 +34,39 @@
   width: 8px;
 }
 
-.intent-info {
+.fill-info {
   fill: $color-base-blue;
 }
 
-.intent-primary {
+.fill-primary {
   fill: $color-base-purple;
 }
 
-.intent-success {
+.fill-success {
   fill: $color-base-green;
 }
 
-.intent-danger {
+.fill-danger {
   fill: $color-base-red;
 }
 
-.intent-warning {
+.fill-warning {
   fill: $color-base-orange;
 }
 
-.intent-flag {
+.fill-inverted {
   fill: $color-base-white;
 }
 
+.fill-default {
+  fill: $color-core-neutral-7;
+}
 
+.fill-disabled {
+  fill: $color-code-neutral-5;
+
+  &-light {
+    fill: $color-code-neutral-4;
+  }
+}
 
--- a/packages/ui-components/src/Icon/Icon.module.scss
+++ b/packages/ui-components/src/Icon/Icon.module.scss
@@ -23,6 +23,7 @@
   width: 20px;
 }
 
+.size-default,
 .size-small {
   height: 16px;
   width: 16px;
@@ -38,7 +39,7 @@
 }
 
 .intent-primary {
-  fill: $color-base-blue;
+  fill: $color-base-purple;
 }
 
 .intent-success {
--- a/packages/ui-components/src/Icon/Icon.module.scss
+++ b/packages/ui-components/src/Icon/Icon.module.scss
@@ -1,6 +1,6 @@
 @import "../styles/tokens.scss";
 .icon {
-  fill: $crl-base-text--dark;
+  fill: $color-core-neutral-7;
   margin: auto;
 
   &:disabled {
--- a/packages/ui-components/src/Icon/Icon.module.scss
+++ b/packages/ui-components/src/Icon/Icon.module.scss
@@ -6,40 +6,31 @@
   &:disabled {
     pointer-events: none;
   }
-
-  &__container {
-    display: inline-flex;
-  }
-}
-
-.size-xx-large {
-  height: 80px;
-  width: 80px;
 }
 
 .size-x-large {
-  height: 56px;
-  width: 56px;
-}
-
-.size-large {
-  height: 48px;
-  width: 48px;
-}
-
-.size-medium {
   height: 32px;
   width: 32px;
 }
 
-.size-default {
-  height: 46px;
+.size-large {
+  height: 24px;
   width: 24px;
+}
+
+.size-medium {
+  height: 20px;
+  width: 20px;
 }
 
 .size-small {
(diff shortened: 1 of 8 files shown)


packages/ui-components/src/Icon/Icon.module.scss
```diff
@@ -1,12 +1,15 @@
...
 .icon {
-  display: inline-flex;
+  fill: $crl-base-text--dark;
   margin: auto;
 
   &:disabled {
     pointer-events: none;
   }
+
+  &__container {
```
nathanstilwell, line 10: The Icon component should not have a container. A container can be provided by whoever is implementing it. The component should only have considerations for the SVG it is rendering. We want this component to consider itself like an image.

packages/ui-components/src/Icon/Icon.module.scss
```diff
@@ -30,7 +33,7 @@
 }
 
 .size-default {
-  height: 24px;
+  height: 46px;
```
nathanstilwell, line 36: This height isn't specified in the design at all. This feels like a specific number for use in a context (like an Input) and doesn't belong here.

packages/ui-components/src/Icon/Icon.tsx
```diff
@@ -15,45 +15,51 @@ export type IconSize =
...
 
-export type IconTint =
-  | "blue"
-  | "green"
-  | "red"
-  | "white"
-  | "neutral"
-  | "inherit"
-  | "orange";
+export type IconStyle =
```
nathanstilwell, line 18: For consistency we should call this IconIntent
```suggestion
export type IconIntent =
```

packages/ui-components/src/Icon/Icon.tsx
```diff
@@ -15,45 +15,51 @@ export type IconSize =
...
+
   const Element = get(Icons, iconName, null);
 
   if (Element === null) {
     return null;
   }
 
-  return <Element className={classnames} {...props} />;
+  return (
+  <div className="icon__container">
```
nathanstilwell, line 60: As previously noted, this breaks the modularity of the component.

nathanstilwell (changes requested):
Wrapping the Icon in a container makes it a lot less flexible. This component should behave like an SVG or a image element and leave any modification of display or position to the implementing application or component.

packages/ui-components/src/Icon/Icon.module.scss
poojamaniar: Alright, makes sense. Updated.

packages/ui-components/src/Icon/Icon.module.scss
poojamaniar: yup, updated these values from figma so they reflect latest design icon sizes.

packages/ui-components/src/Icon/Icon.module.scss
```diff
@@ -1,68 +1,61 @@
 @import "../styles/tokens.scss";
-
 .icon {
-  display: inline-flex;
```
poojamaniar, line 4: This was removed because this assumes icon will be rendered with another element for it to be inline with something else -> sine we want the Icon component to resemble an image, we should be passing this attribute to whatever renders the Icon component.

Has been tested.

packages/ui-components/src/Icon/Icon.module.scss
```diff
@@ -1,68 +1,61 @@
...
 
   &:disabled {
     pointer-events: none;
   }
 }
 
-.size-xx-large {
-  height: 80px;
-  width: 80px;
-}
```
poojamaniar, line 15: xx large is longer defined in Figma

packages/ui-components/src/Icon/Icon.module.scss
nathanstilwell: 👍🏼

packages/ui-components/src/Icon/Icon.module.scss
```diff
@@ -1,68 +1,62 @@
...
+.intent-danger {
   fill: $color-base-red;
 }
 
-.tint-orange {
+.intent-warning {
   fill: $color-base-orange;
 }
 
-.tint-white {
```
nathanstilwell, line 58: Is this one meant to be the "Knockout" color in Figma (when rendered on a dark background)?

packages/ui-components/src/Icon/Icon.tsx
```diff
@@ -8,45 +8,52 @@ import styles from "./Icon.module.scss";
...
+
+export type IconIntent =
+  | "danger"
+  | "default"
+  | "info"
+  | "primary"
+  | "success"
+  | "warning";
+
+// The fill of an icon can be:
```
nathanstilwell, line 26: This is a good comment.

packages/ui-components/src/Icon/Icon.tsx
```diff
@@ -8,45 +8,52 @@ import styles from "./Icon.module.scss";
...
+//      inverted (when the icon appears against a dark background)
+//      disabled (for when the background is neutral-2)
+//      disabled-light (for when the background is neutral-0)
+export type IconFill = IconIntent | "inverted" | "disabled" | "disabled-light";
 
 type OwnIconProps = {
   iconName: keyof typeof Icons;
   size?: IconSize;
-  tint?: IconTint;
+  intent?: IconIntent;
```
nathanstilwell, line 36: this should be `fill` now instead of `intent`.

packages/ui-components/src/InlineAlert/InlineAlert.tsx
```diff
@@ -2,14 +2,12 @@ import React, { useMemo } from "react";
...
-import { Icon, IconTint } from "../Icon/Icon";
-
-export type InlineAlertIntent = "info" | "error" | "success" | "warning";
+import { Icon, IconIntent } from "../Icon/Icon";
 
 const cx = classNames.bind(styles);
 
 interface InlineAlertBaseProps {
-  intent?: InlineAlertIntent;
+  intent?: IconIntent;
```
nathanstilwell, line 10: I don't mind this here for now, but this makes me think long term we may need a type defined for Intents that are not component specific. That would leveraged by components like you've done here.

nathanstilwell (changes requested):
I think there's only one small issue remaining, but this is looking really good!

packages/ui-components/src/InlineAlert/InlineAlert.tsx
poojamaniar: Yes! I can make an issue for it in Github for ui-components.

nathanstilwell (approved):
LGTM

PR #145, opened 2020-12-05
# icons, ui-components // Icon Updates

nathanstilwell:
# Icon Updates

## Overview

Updating some system icons (the original set contained in the @cockroachlabs/icons package) and adding additional sets from design specification. To accommodate these new sets, new components will be created.

## Updated Icons
Screen Shot 2020-12-10 at 10 15 25

As we separating icon sets, internally we are referring to this original icon set as "System icons". The default fill color changed in the exported images (from `black` to `#242A35`) to be more comfortable to look at in the sample. This can be **safely ignored in the review** as this default `fill` attribute is stripped in the conversion to a React component to be overridden by a prop. 

Screen Shot 2020-12-10 at 10 10 40

Added new icons to the System UI icons set; `Gear`, `Open`, and `Star`

Screen Shot 2020-12-10 at 10 14 45

Removed the `Loading` Icon as this has become an independant component ([`Spinner`](https://github.com/cockroachdb/ui/tree/master/packages/ui-components/src/Spinner))

## New Icon Sets
Outside of the System icons, there are other SVG images we use as icons and images throughout the UI. Each of these sets have slightly different treatments (sizes and fills) and so were separated into logical groupings around treatments or content.
 
Screen Shot 2020-12-10 at 10 27 08
Screen Shot 2020-12-10 at 10 27 21
Screen Shot 2020-12-10 at 10 27 51
Screen Shot 2020-12-10 at 10 28 05
Screen Shot 2020-12-10 at 10 28 23

The `index.ts` of the `icons` package was updated to reflect this change, by exporting these sets as a named group (`SystemIcons`, `Pictograms`, `ThirdParty`, etc). We are also continuing to export each Icon by name to be backward compatible with current usage. For example,

```ts
// import a single icon
import { Download } from "@cockroachlabs/icons";

// import a set
import { Pictograms } from "@cockroachlabs/icons";
```

## Preserving `fill` attribute
Some icons (third party logo, flags, etc) need to have their `fill` attributes preserved and are not candidate for having intent colors applied to them, while other icons (System and Pictograms) do. To accommodate both needs, we have separated the `svg` directory into two sub-directories for each purpose; `customFill` and `preserveFill`. Each sub-directory will be processed independently by the build step with the `customFill` directory using the SVGO config to remote fill attributes and `preserveFill` suppressing this config. When future SVGs are added, they should be added to one of these two sub-directories.

## A Component per set
To encapsulate the treatments of the individual icon sets, we have created a component per set and exposed treatment options as props. The new components introduced in this PR are,
- `Pictogram`
- `ThirdPartyIcon`
- `CreditCard`
- `Flag`
- `Illustration`


packages/icons/src/index.ts
```diff
@@ -1,44 +1,294 @@
...
+import { default as Menu } from "./components/Menu";
+import { default as Nodes } from "./components/Nodes";
+import { default as NotFound } from "./components/NotFound";
+import { default as SleepyMoon } from "./components/SleepyMoon";
+
+// 3rd Party Icons
+import { default as Aws } from "./components/Aws";
+import { default as Gcp } from "./components/Gcp";
+
+// Cards and Flags
```
laurenbarker, line 80: nit: `s/Cards and Flags/Cards/g`

packages/icons/src/index.ts
```diff
@@ -1,44 +1,294 @@
...
+import { default as Monitoring } from "./components/Monitoring";
+
+// Illustrations
+import { default as MagnifyingGlass } from "./components/MagnifyingGlass";
+import { default as Menu } from "./components/Menu";
+import { default as Nodes } from "./components/Nodes";
+import { default as NotFound } from "./components/NotFound";
+import { default as SleepyMoon } from "./components/SleepyMoon";
+
+// 3rd Party Icons
```
laurenbarker, line 76: I don't consider this blocking, but it's worth noting that these icons are only used in CC Console. Having a separate icon set in CC Console for CC Console specific icons would make it easier for us to modify or add to the existing set. If they are in the UI repository, then we will need to update multiple packages and bump the versions of each in order to add anything. It's a lot of friction.

I believe that 3rd party icons, cards, and flags are only used in CC Console.

packages/icons/src/index.ts
nathanstilwell: 👍🏼

packages/icons/src/index.ts
nathanstilwell: While it is true that currently the majority of these icons are only used in the CockroachCloud console, they are a universally consistent part of the design system. So when we use flags or credit  cards or third party logos, we are using consistent graphics. I'm not sure if we have Cloud console _specific_ icons, but if we do please let me know.

I am sympathetic to the potentially increased friction of updating svgs outside of an application repository, but historically there have been very few changes to these icons once they exists. If the concern is that having icons centralized will hinder development at the application level, I would point out that engineers are free to do what they need to do to meet their goals. If that means using local SVGs or local icon components over centralized ones, we should discuss that as those needs arise. But the goal of this repository remains to provide portable resources that are consistent with Design System specification and I believe these icon sets are included in that.

Annebirzin:
@nathanstilwell attached here are the credit card icon svgs so you have them (I slightly tweaked the visa icon here to make sure the sizes are all the same) 
[credit-card-icons.zip](https://github.com/cockroachdb/ui/files/5681304/credit-card-icons.zip)

Annebirzin:
@nathanstilwell I noticed that a few of the flag, pictograms and illustrations didn't have the proper fills coming through from the svg figma export (attached screenshot of 'add' and 'location'). 

Screen Shot 2020-12-14 at 2 10 48 PM

I've updated the following icons and attached the svgs here:

- flags: canada, france, ireland, japan, korea, mumbai, singapore
- pictograms: add, location, monitoring
- illustrations: sleepymoon

[icon-updates.zip](https://github.com/cockroachdb/ui/files/5691050/icon-updates.zip)

packages/ui-components/src/Icon/Illustration.module.scss
```diff
@@ -0,0 +1,4 @@
+.illustration {
+  height: 152px;
```
koorosh, line 2: @nathanstilwell , can it be useful to make illustrations more responsive? My concern is about cases when page is resized to smaller sizes. I would suggest 
- to use maxHeight and maxWidth 
- or allow illustrations to fill entire parent's container with preserved image proportions

koorosh (approved):
Looks great! Only one question I have is about hard coded height/width sizes for Illustration component

packages/ui-components/src/Icon/Illustration.module.scss
nathanstilwell: That's a good idea. We might as well use the responsive nature of these SVGs. Given where we use these illustrations (usually in the middle of an empty page or section) I don't believe we would see this issue very often, but it's good to be forward thinking.

PR #148, opened 2020-12-11
Admin UI components migration

dhartunian:
# Admin UI Components // Initial commit with port of git history

Used the following command to move repo over:
```
git subtree add --prefix=packages/admin-ui-components ../admin-ui-components/ master
```

Had to do a few tweaks in separate commit after porting history:
- Move `admin-ui-components` publish workflow to the top level github workflows directory.

- Edit `admin-ui-components publish workflow to only run inside package subdirectory.

- Update versions to match what's currently in the packages in `admin-ui-components` package.json

- Minor fix in `loading.tsx` after version bump

#### Checklist
- [X] I have written or updated test for the changes I made (repo came w/ tests)
- [X] I have updated the README of the package I'm working in to reflect my changes (repo came w/ README)
- [X] I have added or updated Storybook if appropriate for my changes (repo storybook runs 👍🏻)


dhartunian:
Reviewer notes: Please take a look at the final commit in the chain which is the only one w/ work to make this port work. All remaining commits are the repo history from `admin-ui-components`

.github/workflows/publish-admin-ui-components.yml
```diff
@@ -30,19 +34,24 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
       - name: Install dependencies
         run: yarn install
```
nathanstilwell, line 36: Here I think we could use `yarn install --production` since at this point we don't need any dev dependencies. (I just noticed this same thing for `ui-components`)

dhartunian: done!

nathanstilwell (approved):
LGTM!

PR #154, opened 2020-12-15
Admin UI components rename

dhartunian:
Admin UI Components // Rename Admin UI Components to Cluster UI

This PR changes the name of the "admin-ui-components" 
project to "cluster-ui".

Code changes were made where appropriate to reflect the
change.

Omitted from this change are:
- any CSS modifications
- any changes to Docs links and URLs since those are 
  managed separately

#### Checklist
- [X] I have written or updated test for the changes I made -- N/A
- [X] I have updated the README of the package I'm working in to reflect my changes
- [X] I have added or updated Storybook if appropriate for my changes -- N/A

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/ui/154)


dhartunian:
This PR will wait for https://github.com/cockroachdb/ui/pull/148 to get merged first, then this won't look like such a gigantic change.

nathanstilwell:
I suppose my first comment is about the name "DB Console Components". I that what this package is for? To put components in DB Console? Or rather to expose CC Console to views present in DB Console? But isn't there some other categorization that is true for the components of this package? Will this package ever contain a component that didn't get extracted from DB Console?

dhartunian:
> I suppose my first comment is about the name "DB Console Components". I that what this package is for? To put components in DB Console? Or rather to expose CC Console to views present in DB Console? But isn't there some other categorization that is true for the components of this package? Will this package ever contain a component that didn't get extracted from DB Console?

I see this package as containing DB Console "features" in terms of large-scope components. The Transactions component was, for instance, built in this repo first instead of being extracted from CRDB. So maybe to pull away from the "components" terminology we could say `db-console-features`. What do you think of that? That kinda widens the scope instead of narrowing but it might be more accurate.

nathanstilwell:
> So maybe to pull away from the "components" terminology we could say `db-console-features`. 

I like this suggestion, as these large _page sized views_ are only _components_ on a technicality. I think "features" is a better than components.

dhartunian:
@nathanstilwell @nkodali Following up about the discussion we had on this, I was thinking we could use "cluster-ui" for the package to decouple from the "db console" name and any future changes and to underscore the fact that this package is *not* a components package.

I remember us discussing that one and I like that it's short.

nathanstilwell:
I think `cluster-ui` is as good as it is going to get for now. I like it a lot actually. Let me know if I can help with this.

dhartunian:
@nathanstilwell RFAL

nathanstilwell (approved):
LGTM!

PR #161, opened 2020-12-17
db-console: directly set base path for API requests

koorosh:
# db-console // provide base path for API requests

Ported from PR https://github.com/cockroachdb/admin-ui-components/pull/48

#### Checklist

With new requirements, base path for api calls can be arbitrary and
might not rely on browser's path or whatever.
This change implements getter/setter functions to directly set
base path which is needed to use for api requests.

Also it simplifies and eliminates passing basePath as extra
argument through sagas.


dhartunian (approved):
LGTM

PR #162, opened 2020-12-17
db-console: statement details component

koorosh:
# db-console // StatementDetails component

Related to https://github.com/cockroachdb/cockroach/issues/57372

This change extracts Statements details page and its dependent
components and styles.
- most of the components moved without any changes, only import
paths were adjusted
- styles converted from Styl to SCSS
- modified route paths to ensure they start with "/" root path
(it ensures that routes behave the same way regardless to current
path).
- `util` directory now contain `network` and `nodes` subdirectories
with logic specific to particular entities. It allows avoid extra logic
in redux layer and keep it independently.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/ui/162)


packages/admin-ui-components/src/loading/loading.tsx
```diff
@@ -99,7 +99,7 @@ export const Loading: React.FC<LoadingProps> = props => {
 
     return (
       <div className={cx("alerts-container", props.errorClassName)}>
-        {errorAlerts}
+        {React.Children.toArray(errorAlerts)}
```
dhartunian, line 102: why is this necessary?

packages/admin-ui-components/webpack.config.js
```diff
@@ -21,7 +21,7 @@ module.exports = {
       'node_modules',
       path.join(__dirname, 'src/fonts'),
     ],
-    extensions: [".ts", ".tsx", ".js", ".jsx", ".less"],
+    extensions: [".ts", ".tsx", ".js", ".jsx", ".less", ".scss"],
```
dhartunian, line 24: why did we need to add this now?

dhartunian (changes requested):
Just a few minor questions from me.

I don't like having a `util` directory that accumulates lots of stuff. Could we put the `nodes` and `network` code just in `src`?

packages/admin-ui-components/src/loading/loading.tsx
koorosh: It fixes React error about missing `key` attribute for array of elements.
`React.Children.toArray` accepts an array of react elements and allows to set automatically `key` attribute if it's missed.

packages/admin-ui-components/webpack.config.js
koorosh: hm, this change looks useless, will revert it.

koorosh:
> I don't like having a `util` directory that accumulates lots of stuff. Could we put the `nodes` and `network` code just in `src`?

Agree

packages/admin-ui-components/src/barCharts/index.ts
```diff
@@ -1 +1,5 @@
 export * from "./barCharts";
+export * from "./rowsBrealdown";
```
elkmaster, line 2: typo

packages/admin-ui-components/webpack.config.js
```diff
@@ -98,6 +98,7 @@ module.exports = {
         test: /\.js$/,
         loader: "source-map-loader",
       },
+      { test: /\.css$/, use: [ "style-loader", "css-loader" ] },
```
dhartunian, line 101: Where are we loading raw CSS?

dhartunian (approved):
LGTM, just one small question.

packages/admin-ui-components/webpack.config.js
koorosh: Components which depend on Antd styles, have to load its CSS styles explicitly and we have Table component which depends on it:
https://github.com/cockroachdb/ui/pull/162/files#diff-70a6adf902647bd71cee4e2b65f55a1a999736c39698aef2ba64f1a70aedbadaR6

elkmaster (approved):

PR #164, opened 2020-12-17
ui: Extract Sessions Page from DB Console

elkmaster:
ui: Extract Sessions Page from DB Console

extract session page into ui repo:
- most of the components moved without any or minimal changes
- styles converted to scss modules

Resolves: #57296

Release note (ui): none

### DIFF
--- a/packages/admin-ui-components/src/statementsTable/statementsTableContent.tsx
+++ b/packages/admin-ui-components/src/statementsTable/statementsTableContent.tsx
@@ -276,19 +276,35 @@ export const StatementTableCell = {
   ),
 };
 
-export const StatementLink = (props: {
+interface StatementLinkProps {
   statement: string;
   app: string;
   implicitTxn: boolean;
   search: string;
-}) => {
+  anonStatement?: string;
+}
+
+// StatementLinkTarget returns the link to the relevant statement page, given
+// the input statement details.
+export const StatementLinkTarget = (props: StatementLinkProps) => {
+  let base: string;
+  if (props.app && props.app.length > 0) {
+    base = `/statements/${props.app}/${props.implicitTxn}`;
+  } else {
+    base = `/statement/${props.implicitTxn}`;
+  }
+
+  let linkStatement = props.statement;
+  if (props.anonStatement) {
+    linkStatement = props.anonStatement;
+  }
+  return `${base}/${encodeURIComponent(linkStatement)}`;
+};
+
+export const StatementLink = (props: StatementLinkProps) => {
   const summary = summarize(props.statement);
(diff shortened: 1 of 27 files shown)


elkmaster:
corresponding pr for admin-ui https://github.com/cockroachdb/cockroach/pull/58034

packages/admin-ui-components/src/tooltip2/tooltip.tsx
```diff
@@ -9,19 +9,32 @@ import styles from "./tooltip.module.scss";
...
 export const Tooltip = (props: TooltipProps & AntTooltipProps) => {
-  const { children, theme, overlayClassName } = props;
+  const { children, theme, overlayClassName, tableTitle } = props;
   const classes = cx(
     "tooltip-overlay",
     `crl-tooltip--theme-${theme}`,
     overlayClassName,
   );
+
+  const title = tableTitle ? (
```
dhartunian, line 25: I don't see this in the source of `pkg/ui`. Why did we need to add it?

packages/admin-ui-components/src/statementsTable/statementsTableContent.tsx
```diff
@@ -276,19 +276,35 @@ export const StatementTableCell = {
...
   app: string;
   implicitTxn: boolean;
   search: string;
-}) => {
+  anonStatement?: string;
+}
+
+// StatementLinkTarget returns the link to the relevant statement page, given
+// the input statement details.
+export const StatementLinkTarget = (props: StatementLinkProps) => {
```
dhartunian, line 289: Would this introduce problems in CC if we need to route to a different URL set?

dhartunian (changes requested):
Just minor questions, otherwise looks good.

packages/admin-ui-components/src/tooltip2/tooltip.tsx
elkmaster: we have different style for table title tooltip. to achieve that earlier we added <div/> with corresponding class(https://github.com/cockroachdb/cockroach/blob/master/pkg/ui/src/views/sessions/sessionsTableContent.tsx), but to remove duplications I moved this on tooltip component level.

packages/admin-ui-components/src/statementsTable/statementsTableContent.tsx
elkmaster: Looks like it should not. Current change is just moving out basically the same url forming logic to separate function. But I will double check it when will be working on CC integration/

dhartunian (approved):

PR #165, opened 2020-12-21
[DS-13] Form update: edits from forms retro

poojamaniar:
# ui-components // edits from forms retro (re-type password field, reveal prop, type changes in Field component, multiple validators for password)

[Jira Issue](https://cockroachlabs.atlassian.net/browse/DS-13
)
[Forms Retro](https://docs.google.com/document/d/1e83kWWWck9FsVcM8HyShvH6dJ8RLG8UfcyrIy4Bi9kw/edit?ts=5fdbcf71
)

GIF of Password Changes (reveal prop, Re-type password field, validation):

Screenshot of multiple validators:
image

#### Checklist

### DIFF
--- a/packages/ui-components/src/Input/CheckboxInput.tsx
+++ b/packages/ui-components/src/Input/CheckboxInput.tsx
@@ -35,9 +35,6 @@ export const CheckboxInput = ({
     <CommonInput
       className={className}
       error={error}
-      invalid={invalid}
-      label={label}
-      ariaLabel={ariaLabel}
       {...rest}
       fieldInput={fieldInput}
     />
--- a/packages/ui-components/src/Input/CheckboxInput.tsx
+++ b/packages/ui-components/src/Input/CheckboxInput.tsx
@@ -2,10 +2,6 @@ import React from "react";
 import classNames from "classnames/bind";
 import { CommonInputProps, CommonInput } from "./CommonInput";
 
-type CheckboxProps = React.InputHTMLAttributes<HTMLInputElement>;
-
-export type CheckboxInputProps = CommonInputProps & CheckboxProps;
-
 export const CheckboxInput = ({
   id,
   className,
@@ -15,7 +11,7 @@ export const CheckboxInput = ({
   ariaLabel,
   required,
   ...rest
-}: CheckboxInputProps) => {
+}: CommonInputProps) => {
   const inputProps = {
     id: id,
     className: classNames("crl-input", className, {
--- a/packages/ui-components/src/Input/CommonInput.tsx
+++ b/packages/ui-components/src/Input/CommonInput.tsx
@@ -1,26 +1,28 @@
 import React from "react";
 import classNames from "classnames/bind";
 import "./input.module.scss";
-import { FieldMetaState } from "react-final-form";
 
-export interface CommonInputProps {
+export interface CommonInputProps
+  extends React.DetailedHTMLProps<
+    React.InputHTMLAttributes<HTMLInputElement>,
+    HTMLInputElement
+  > {
   // this should be the element containing the input
   fieldInput?: JSX.Element;
-  className?: string;
   help?: string | JSX.Element;
   // derived from final form type
-  // error is assigned a value of meta.error
-  // which has type any
+  // error is assigned a value of meta.error which has type any
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   error?: any;
   inline?: boolean;
-  // FieldValue passed to meta is string for all TextTypeInputs
-  // boolean for CheckboxInput, and number for NumberInput
-  meta?: FieldMetaState<string | number | boolean>;
   // following are used in Checkbox, TextType Inputs, which render CommonInput
   invalid?: boolean;
   label?: string | JSX.Element;
+  // these props are used for accessibility and aren't part of DetailedHTMLProps
+  // because they're formatted with dashes (i.e. aria-label)
+  // recall objects don't accept a dash in keys (passed to inputProps in [])
   ariaLabel?: string;
+  ariaLabelledBy?: string;
 }
 
 // a component defining how error and help messages
--- a/packages/ui-components/src/Input/CommonInput.tsx
+++ b/packages/ui-components/src/Input/CommonInput.tsx
@@ -43,15 +43,17 @@ export const CommonInput = ({
 
   return (
     <div
-      className={classNames("input-container", className, {
+      className={classNames(className, {
         inline: inline,
       })}
     >
       {fieldInput}
-      <div className="message">
-        {errorMsg}
-        {helpMsg}
-      </div>
+      {(errorMsg || helpMsg) && (
+        <div className="message">
+          {errorMsg}
+          {helpMsg}
+        </div>
+      )}
     </div>
   );
 };
--- a/packages/ui-components/src/Input/CommonInput.tsx
+++ b/packages/ui-components/src/Input/CommonInput.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 import classNames from "classnames/bind";
 import "./input.module.scss";
-import { FieldRenderProps, FieldMetaState } from "react-final-form";
+import { FieldMetaState } from "react-final-form";
 
 export interface CommonInputProps {
   // this should be the element containing the input
--- a/packages/ui-components/src/Input/CommonInput.tsx
+++ b/packages/ui-components/src/Input/CommonInput.tsx
@@ -10,14 +10,13 @@ export interface CommonInputProps
   // this should be the element containing the input
   fieldInput?: JSX.Element;
   help?: string | JSX.Element;
-  // derived from final form type
-  // error is assigned a value of meta.error which has type any
+  // error is assigned a value of meta.error which has type any in react-final-forms
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   error?: any;
   inline?: boolean;
   // following are used in Checkbox, TextType Inputs, which render CommonInput
   invalid?: boolean;
-  label?: string | JSX.Element;
+  label?: string;
   // these props are used for accessibility and aren't part of DetailedHTMLProps
   // because they're formatted with dashes (i.e. aria-label)
   // recall objects don't accept a dash in keys (passed to inputProps in [])
--- a/packages/ui-components/src/Input/CommonInput.tsx
+++ b/packages/ui-components/src/Input/CommonInput.tsx
@@ -16,7 +16,7 @@ export interface CommonInputProps
   inline?: boolean;
   // following are used in Checkbox, TextType Inputs, which render CommonInput
   invalid?: boolean;
-  label?: string;
+  label?: string | JSX.Element;
   // these props are used for accessibility and aren't part of DetailedHTMLProps
   // because they're formatted with dashes (i.e. aria-label)
   // recall objects don't accept a dash in keys (passed to inputProps in [])
--- a/packages/ui-components/src/Input/EmailPassword.module.scss
+++ b/packages/ui-components/src/Input/EmailPassword.module.scss
@@ -6,9 +6,7 @@
 }
 
 .new-password-validation-container {
-  display: flex;
-  flex-flow: column wrap;
-  list-style: none;
+  column-count: 2;
   margin-bottom: crl-gutters(3);
   padding: 0;
 }
@@ -19,6 +17,18 @@
   font-size: 12px;
 }
 
-.new-password-validation-label {
+.new-password-validation-icon {
+  padding-top: crl-gutters(0.5);
+  display: inline-flex;
+  opacity: 0.35,
+}
+
+.new-password-validation-error,
+.new-password-validation-success {
   padding-left: crl-gutters(0.75);
+  color: $color-font-2;
+}
+
+.new-password-validation-error {
+  color: $color-font-4;
 }
--- a/packages/ui-components/src/Input/EmailPasswordInput.tsx
+++ b/packages/ui-components/src/Input/EmailPasswordInput.tsx
@@ -52,23 +52,21 @@ export const PasswordInput = ({
 
   // we only want to show the check icon if the user is retyping the password
   // and the password matches the previous password field
-  const doesRetypedPasswordMatch = !error && meta && meta.touched &&  (
-    <Icon
-      iconName={"Check"}
-      size="medium"
-      fill="success"
-    />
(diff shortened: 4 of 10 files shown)


packages/ui-components/src/Input/Field.tsx
```diff
@@ -1,70 +1,38 @@
...
+  | typeof EmailInput
+  | typeof SingleLineTextInput
+  | typeof NumberInput
+  | typeof CheckboxInput;
+
+// Extends field render props, which will be passed to the Field component in InputField
+export interface InputFieldProps<
+  IComponent extends React.ComponentType<InputTypes>
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+> extends FieldProps<any, FieldRenderProps<any, HTMLElement>, HTMLElement> {
```
poojamaniar, line 19: We have an `any` here because the type we should pass here is the `FieldValue`, which is `string | boolean | number`, for the different kind of Inputs we have. Not sure how to get rid of the any...

packages/ui-components/src/Input/TextTypeInput.tsx
```diff
@@ -57,18 +52,16 @@ export const BaseTextInput: React.FC<InternalTextProps> = props => {
...
   } = props;
 
   const inputProps = {
+    id: id,
     className: classNames("crl-input", className, {
       "crl-input--prefix": prefixElement,
       "crl-input--suffix": suffix,
       invalid: error || invalid,
     }),
-    name: name,
```
poojamaniar, line 71: We don't destructure name as part of the props, so we can pass it as part of the rest on line 70.

packages/ui-components/src/Input/TextTypeInput.tsx
```diff
@@ -95,18 +88,9 @@ export const BaseTextInput: React.FC<InternalTextProps> = props => {
...
 
-  const labelElement = existingPassword ? (
-    <div className="existing-password-label">
-      {labelDiv}
-      {forgotPasswordLinkElement}
-    </div>
-  ) : (
-    labelDiv
-  );
-
```
poojamaniar, line 106: no longer needed since we don't handle the forgot password link as part of the Input, and instead handle it in CC/DB Console.

packages/ui-components/src/Input/EmailPassword.module.scss
```diff
@@ -6,9 +6,7 @@
 }
 
 .new-password-validation-container {
-  display: flex;
-  flex-flow: column wrap;
-  list-style: none;
+  column-count: 2;
```
poojamaniar, line 9: instead of implementing with column wrap, which will create as many rows as possible before wrapping to the next column, we can enforce a max column number, behaviour is shown in PR description screenshot. Max # of columns confirmed by design on ticket: https://cockroachlabs.atlassian.net/browse/DS-13

packages/ui-components/src/Input/EmailPassword.module.scss
```diff
@@ -19,6 +17,18 @@
...
+}
+
+.new-password-validation-error,
+.new-password-validation-success {
   padding-left: crl-gutters(0.75);
+  color: $color-font-2;
+}
+
+.new-password-validation-error {
+  color: $color-font-4;
```
poojamaniar, line 33: we want to have neutral-7 for success validation messages, and neutral-5 for error validation messages as shown in Figma (https://www.figma.com/file/B5AtEGdRbW3VxiBdg62TB8Xw/CRDB-Design-System?node-id=22%3A4).

packages/ui-components/src/Input/EmailPassword.module.scss
```diff
@@ -19,6 +17,18 @@
   font-size: 12px;
 }
 
-.new-password-validation-label {
+.new-password-validation-icon {
+  padding-top: crl-gutters(0.5);
+  display: inline-flex;
+  opacity: 0.35,
```
poojamaniar, line 23: icons should have an opacity of 0.35 accord to figma

packages/ui-components/src/Input/Field.tsx
```diff
@@ -1,70 +1,38 @@
...
+  as: IComponent;
+  // should include types currently supported
+  type: "text" | "email" | "password" | "number" | "checkbox";
```
poojamaniar, line 22: Wanted to highlight this change, as it requires an additional prop to be passed whereas the field architecture prior (with one field per input - EmailField/PasswordField/TextInput etc) didn't require this. The additional prop is `type`: I've made `type` mandatory as well as the `as` prop, so when a developer wants to include a password field they'd do so `<InputField as={EmailInput} type="email" ...>`. Is this OK?

I believe this is necessary for the InputField to be able to support different Inputs. I tried doing some type checking with `as`, to see if we can pass the `type` prop ourselves depending on what `as` is, but I think it's not as safe as the developer passing it in themselves.

packages/ui-components/src/Input/TextTypeInput.tsx
```diff
@@ -3,13 +3,14 @@ import classNames from "classnames/bind";
...
 import { isEmpty } from "lodash";
 import "./input.module.scss";
+import { FieldRenderProps } from "react-final-form";
 
-export interface TextAndNumberProps
-  extends React.InputHTMLAttributes<HTMLInputElement> {
+// extending the fieldrender props here ensures that we have access to meta and input
+// if we need thems
+export interface TextAndNumberProps<T>
+  extends Partial<FieldRenderProps<T, HTMLInputElement>> {
```
poojamaniar, line 11: Partial because we don't want the field props (input, meta) to be mandatory props for the Text input.

packages/ui-components/src/Input/TextTypeInput.tsx
```diff
@@ -24,26 +25,11 @@ export interface MultilineProps
...
-// the following props are used to implement {Email, Password, Number}Input
-// it is recommended to use those components instead of passing an "email" type to TextInput
-interface CustomProps extends React.InputHTMLAttributes<HTMLInputElement> {
-  ariaLabel?: string;
-  ariaLabelledBy?: string;
-  forgotPasswordLinkElement?: JSX.Element;
-  // this prop is only used for internal implementation
-  // whether the password exists or not, should be indicated by New/Existing PasswordInput
-  existingPassword?: boolean;
-}
```
poojamaniar, line 36: Moved to the `./EmailPassword` file as components have been refactored such that these props are only necessary for `PasswordInput`.

packages/ui-components/src/Input/TextTypeInput.tsx
```diff
@@ -57,18 +43,18 @@ export const BaseTextInput: React.FC<InternalTextProps> = props => {
     type = "text",
     prefixElement,
     ariaLabelledBy,
-    existingPassword = false,
-    forgotPasswordLinkElement,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    inline,
```
poojamaniar, line 47: De-structured because we shouldn't be passing this to `inputProps`, as it causes a noisy error since we're trying to pass an unrecognized prop to the DOM input element. Open to other suggestions of implementation - seems the alternative is to destructure all `inputProps`, which runs into the issue we had before of having to pass `onChange`,`name`, `onFocus`, etc.

packages/ui-components/src/Input/CheckboxInput.tsx
```diff
@@ -2,10 +2,6 @@ import React from "react";
 import classNames from "classnames/bind";
 import { CommonInputProps, CommonInput } from "./CommonInput";
 
-type CheckboxProps = React.InputHTMLAttributes<HTMLInputElement>;
-
-export type CheckboxInputProps = CommonInputProps & CheckboxProps;
-
```
poojamaniar, line 8: `CommonInputProps` now extends `InputHTMLAttributes`, so CheckboxInput only needs to extend `CommonInputProps`

packages/ui-components/src/Input/TextTypeInput.tsx
poojamaniar: also, `T` is used here as we need to specify whether a `text/number` input is being used (this type refers to `FieldValue`, which should be specified instead of having a `text | number` type

packages/ui-components/src/Input/TextTypeInput.tsx
```diff
@@ -3,13 +3,14 @@ import classNames from "classnames/bind";
...
-export interface TextAndNumberProps
-  extends React.InputHTMLAttributes<HTMLInputElement> {
+// extending the fieldrender props here ensures that we have access to meta and input
+// if we need thems
+export interface TextAndNumberProps<T>
+  extends Partial<FieldRenderProps<T, HTMLInputElement>> {
   suffix?: JSX.Element;
   prefixElement?: JSX.Element;
-  // prop only used internally for implementation
-  type?: "text" | "email" | "password" | "number";
```
poojamaniar, line 12: moved to Field.tsx

packages/ui-components/src/Input/CommonInput.tsx
```diff
@@ -1,26 +1,28 @@
...
   // derived from final form type
-  // error is assigned a value of meta.error
-  // which has type any
+  // error is assigned a value of meta.error which has type any
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   error?: any;
   inline?: boolean;
-  // FieldValue passed to meta is string for all TextTypeInputs
-  // boolean for CheckboxInput, and number for NumberInput
-  meta?: FieldMetaState<string | number | boolean>;
```
poojamaniar, line 19: Handled by `TextAndNumberProps`, since that extends FieldRenderProps, which defines meta and input

packages/ui-components/src/Input/CheckboxInput.tsx
```diff
@@ -39,9 +35,6 @@ export const CheckboxInput = ({
     <CommonInput
       className={className}
       error={error}
-      invalid={invalid}
-      label={label}
-      ariaLabel={ariaLabel}
```
poojamaniar, line 44: these props are not needed or used in CommonInput

packages/ui-components/src/Input/CommonInput.tsx
```diff
@@ -41,15 +43,17 @@ export const CommonInput = ({
...
-      <div className="message">
-        {errorMsg}
-        {helpMsg}
-      </div>
+      {(errorMsg || helpMsg) && (
+        <div className="message">
+          {errorMsg}
+          {helpMsg}
+        </div>
+      )}
```
poojamaniar, line 56: to match the behaviour in Figma, only show `padding` under the `input` if there is a `help/error message`

packages/ui-components/src/Input/TextTypeInput.tsx
poojamaniar: Nevermind, as per design in [DS-13](https://cockroachlabs.atlassian.net/browse/DS-13) forgot password should be displayed within Field

packages/ui-components/src/Input/CommonInput.tsx
```diff
@@ -1,26 +1,27 @@
...
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   error?: any;
   inline?: boolean;
-  // FieldValue passed to meta is string for all TextTypeInputs
-  // boolean for CheckboxInput, and number for NumberInput
-  meta?: FieldMetaState<string | number | boolean>;
   // following are used in Checkbox, TextType Inputs, which render CommonInput
   invalid?: boolean;
-  label?: string | JSX.Element;
+  label?: string;
```
poojamaniar, line 19: `label` doesn't exist on `DetailedHTMLProps`

packages/ui-components/src/Input/EmailPasswordInput.tsx
```diff
@@ -1,81 +1,137 @@
...
-// since we use at most one validator currently
-// we only want to accept one validator label
-export type NewPasswordProps = Omit<
-  ExistingPasswordProps & { validatorLabel: string },
-  "forgotPasswordLinkElement"
->;
-export type EmailProps = NewPasswordProps;
+export interface PasswordProps
+  extends Partial<FieldRenderProps<"text", HTMLInputElement>>,
+    TextInputProps {
```
poojamaniar, line 14: moved the `FieldRenderProps` (which define `input` and `meta`) here, since the other Input components don't need them

packages/ui-components/src/Input/Field.tsx
```diff
@@ -1,70 +1,42 @@
...
+// remove any properties defined in InputFieldProps, so we can handle them exclusively.
+type FieldPropsRefined<
+  IComponent extends React.ComponentType<InputComponentTypes>
+> = InputFieldProps<IComponent> &
+  Omit<
+    React.ComponentPropsWithoutRef<IComponent>,
+    keyof InputFieldProps<IComponent>
+  >;
+
+export function InputField({
```
nathanstilwell, line 38: Shouldn't this file be named `InputField` as we generally name the file after the default or primary export? Also this should probably be the `default` export.

packages/ui-components/src/Input/Field.tsx
```diff
@@ -1,70 +1,42 @@
...
-  | EmailInputFieldProps
-  | NewPasswordInputFieldProps
-  | ExistingPasswordInputFieldProps;
-
-const InputField = ({
+import { EmailInput, PasswordInput } from "./EmailPasswordInput";
+import { CheckboxInput } from "./CheckboxInput";
+
+type InputComponentTypes =
+  | typeof PasswordInput
```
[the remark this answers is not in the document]
poojamaniar: All of those inputs are `React.FC`, I can replace this

packages/ui-components/src/Input/Field.tsx
poojamaniar: Okay I can rename the file to `InputField`, since we can't name this component `Field` since it renders a child component `Field` as well.

packages/ui-components/src/Input/Field.tsx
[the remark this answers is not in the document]
poojamaniar: On second though, I recall why I did this `typeof` union - it was to ensure that the `props` we get always match one of those `Input`s. The `typeof PasswordInput` is a specific `FC`, with `PasswordInputProps` as the params and `JSX.Element` as the return value. 
That is to say, this is what the type boils down to:
```
type InputComponentTypes =
(({ meta, input, reveal, validatorArray, newPassword, retypePassword, ...rest }: PasswordProps) => JSX.Element) | 
((props: TextInputProps) => JSX.Element)
```
... etc

I will document this in the code - do you think this type work could be further simplified?

PR #166, opened 2020-12-22
transactions page: aggregate by fingerprint 

dhartunian:
This change introduces some helper functions that let us aggregate
transaction stats returned by the DB by the fingerprint and app. When
CRDB returns transaction stats there may be copies of the transaction
across different node_ids which executed them. In the UI we don't
present that distinction in the transaction list so, like we do with
statements, we need to aggregate stats across copies of the same
fingerprint.

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/ui/166)


koorosh (commented):
Very easy to read the code! I left more questions to discuss then requested changes.
Also `src/transactionsPage/utils.ts` file has linting errors which would be good to fix.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/ui/166#-:-MPEogEI7NHRBAuYJFTk:b-9e3bej)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @dhartunian, @elkmaster, @koorosh, @nathanstilwell, and @nkodali)

___
*[packages/admin-ui-components/src/transactionsPage/transactionsPage.tsx, line 197 at r1](https://reviewable.io/reviews/cockroachdb/ui/166#-MPEjykY9qvPRhHyrPev:-MPEjykY9qvPRhHyrPew:beyoyc4) ([raw file](https://github.com/cockroachdb/ui/blob/308731d66c87d05d46e9f11b7b9ee17aed627d75/packages/admin-ui-components/src/transactionsPage/transactionsPage.tsx#L197)):*
> ```XML
>             const { pagination, search, filters } = this.state;
>             const { statements, internal_app_name_prefix } = data;
>             const transactions = aggregateAcrossNodeIDs(data.transactions, statements)
> ```

@dhartunian , could this aggregation cause some performance issues? If yes, it might be a case to use memoization (for instance `_.memoize`)
___
*[packages/admin-ui-components/src/transactionsPage/utils.ts, line 133 at r1](https://reviewable.io/reviews/cockroachdb/ui/166#-MPEnBuV5BFx-wqiqG1c:-MPEnBuV5BFx-wqiqG1d:bxa4ryz) ([raw file](https://github.com/cockroachdb/ui/blob/308731d66c87d05d46e9f11b7b9ee17aed627d75/packages/admin-ui-components/src/transactionsPage/utils.ts#L133)):*
> ```XML
>       : b.max_retries,
>     num_rows: addNumericStats(a.num_rows, b.num_rows, countA, countB),
>     service_lat: addNumericStats(a.service_lat, b.service_lat, countA, countB),
> ```

Is it correct to sum up latencies (service_lat, retry_lat, and commit_lat) to show it as aggregated value?
For instance, having the same transaction executed on 10 different nodes with the same `service_lat` equal to 5ms - we get 50ms as aggregated value, but if we decrease the number of nodes - we get lower latency. 
I wondering if it's proper behavor for both statements and transactions.
___
*[packages/admin-ui-components/src/transactionsPage/utils.ts, line 148 at r1](https://reviewable.io/reviews/cockroachdb/ui/166#-MPEhQAn7ax4Qxf8sm9G:-MPEhQAn7ax4Qxf8sm9H:b811i8h) ([raw file](https://github.com/cockroachdb/ui/blob/308731d66c87d05d46e9f11b7b9ee17aed627d75/packages/admin-ui-components/src/transactionsPage/utils.ts#L148)):*
> ```XML
> // provided
> const mergeTransactionStats = function(txns: Transaction[]): Transaction {
>   const txn = {...txns[0]}
> ```

nit, for safety reasons, would be good to validate that `txns` array isn't empty.
___
*[packages/admin-ui-components/src/transactionsPage/utils.ts, line 165 at r1](https://reviewable.io/reviews/cockroachdb/ui/166#-MPEhkJ_0NdBq4PcEAgl:-MPEhkJ_0NdBq4PcEAgm:b-fv83sp) ([raw file](https://github.com/cockroachdb/ui/blob/308731d66c87d05d46e9f11b7b9ee17aed627d75/packages/admin-ui-components/src/transactionsPage/utils.ts#L165)):*
> ```XML
>   return _.chain(t)
>     .map(t => withFingerprint(t, stmts))
>     .groupBy(t => t.fingerprint + t.stats_data.app)
> ```

@dhartunian , what is the purpose for distinguishing transactions that ran by different apps?
Why can't we assume that it is the same transaction?

dhartunian (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/ui/166)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @elkmaster, @koorosh, @nathanstilwell, and @nkodali)

___
*[packages/admin-ui-components/src/transactionsPage/transactionsPage.tsx, line 197 at r1](https://reviewable.io/reviews/cockroachdb/ui/166#-MPEjykY9qvPRhHyrPev:-MPGF-V201nL1okc5M6o:bb1glte) ([raw file](https://github.com/cockroachdb/ui/blob/308731d66c87d05d46e9f11b7b9ee17aed627d75/packages/admin-ui-components/src/transactionsPage/transactionsPage.tsx#L197)):*
<details><summary><i>Previously, koorosh (Andrii Vorobiov) wrote…</i></summary><blockquote>

@dhartunian , could this aggregation cause some performance issues? If yes, it might be a case to use memoization (for instance `_.memoize`)
</blockquote></details>

if I call from a selector, that would solve the problem too, right? it looks like `createSelector` memoizes automatically. I'd rather do that instead of memoize in the component.
___
*[packages/admin-ui-components/src/transactionsPage/utils.ts, line 133 at r1](https://reviewable.io/reviews/cockroachdb/ui/166#-MPEnBuV5BFx-wqiqG1c:-MPGG02r2GCFIWk4Pa2j:biw9064) ([raw file](https://github.com/cockroachdb/ui/blob/308731d66c87d05d46e9f11b7b9ee17aed627d75/packages/admin-ui-components/src/transactionsPage/utils.ts#L133)):*
<details><summary><i>Previously, koorosh (Andrii Vorobiov) wrote…</i></summary><blockquote>

Is it correct to sum up latencies (service_lat, retry_lat, and commit_lat) to show it as aggregated value?
For instance, having the same transaction executed on 10 different nodes with the same `service_lat` equal to 5ms - we get 50ms as aggregated value, but if we decrease the number of nodes - we get lower latency. 
I wondering if it's proper behavor for both statements and transactions.
</blockquote></details>

this is perhaps a bit confusing from the naming but `addNumericStats` is actually computing a new average based on the values and counts of entries. I'll add a test to demonstrate.
___
*[packages/admin-ui-components/src/transactionsPage/utils.ts, line 148 at r1](https://reviewable.io/reviews/cockroachdb/ui/166#-MPEhQAn7ax4Qxf8sm9G:-MPGGHw0Ez1j4K2ybJlh:bs4pxpt) ([raw file](https://github.com/cockroachdb/ui/blob/308731d66c87d05d46e9f11b7b9ee17aed627d75/packages/admin-ui-components/src/transactionsPage/utils.ts#L148)):*
<details><summary><i>Previously, koorosh (Andrii Vorobiov) wrote…</i></summary><blockquote>

nit, for safety reasons, would be good to validate that `txns` array isn't empty.
</blockquote></details>

:thumbsup: yes. I did not feel good about myself when I wrote that :)
___
*[packages/admin-ui-components/src/transactionsPage/utils.ts, line 165 at r1](https://reviewable.io/reviews/cockroachdb/ui/166#-MPEhkJ_0NdBq4PcEAgl:-MPGIyuLE8mC9bJmeWe3:brq918v) ([raw file](https://github.com/cockroachdb/ui/blob/308731d66c87d05d46e9f11b7b9ee17aed627d75/packages/admin-ui-components/src/transactionsPage/utils.ts#L165)):*
<details><summary><i>Previously, koorosh (Andrii Vorobiov) wrote…</i></summary><blockquote>

@dhartunian , what is the purpose for distinguishing transactions that ran by different apps?
Why can't we assume that it is the same transaction?
</blockquote></details>

Good question. This is a bit nuanced because in the statements selector code, the `app` filter is applied *before* aggregation happens so maybe I should do the same here as well.

dhartunian (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/ui/166)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @elkmaster, @koorosh, @nathanstilwell, and @nkodali)

___
*[packages/admin-ui-components/src/transactionsPage/utils.ts, line 133 at r1](https://reviewable.io/reviews/cockroachdb/ui/166#-MPEnBuV5BFx-wqiqG1c:-MQDu1v5--S-dMRK8C_c:b-zf2g2r) ([raw file](https://github.com/cockroachdb/ui/blob/308731d66c87d05d46e9f11b7b9ee17aed627d75/packages/admin-ui-components/src/transactionsPage/utils.ts#L133)):*
<details><summary><i>Previously, dhartunian (David Hartunian) wrote…</i></summary><blockquote>

this is perhaps a bit confusing from the naming but `addNumericStats` is actually computing a new average based on the values and counts of entries. I'll add a test to demonstrate.
</blockquote></details>

I renamed this function. There are already tests in place for it.

elkmaster (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/ui/166#-:-MQN4xSLACzEHY596e0d:bnfp4nl)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @elkmaster, @koorosh, @nathanstilwell, and @nkodali)

PR #168, opened 2020-12-22
set CODEOWNERS on admin-ui-components

dhartunian:
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/ui/168)

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -15,3 +15,4 @@
 # This should make it easy to add new rules without breaking existing ones.
 
 *   @nathanstilwell @j-low @laurenbarker
+packages/admin-ui-components @dhartunian @nkodali @nathanstilwell


nathanstilwell (approved):
LGTM

nathanstilwell (changes requested):
Actually, I take it back. If I'm reading [this example](https://docs.github.com/en/free-pro-team@latest/github/creating-cloning-and-archiving-repositories/about-code-owners) right, the last match wins so if you put the `admin-ui-components` package after the global owners, then only the admin owners will be

CODEOWNERS
```diff
@@ -14,4 +14,5 @@
 # In each subsection folders are ordered first by depth, then alphabetically.
 # This should make it easy to add new rules without breaking existing ones.
 
+packages/admin-ui-components @dhartunian @nkodali @nathanstilwell
 *   @nathanstilwell @j-low @laurenbarker
```
nathanstilwell, line 18:
```suggestion
*   @nathanstilwell @j-low @laurenbarker
packages/admin-ui-components @dhartunian @nkodali @nathanstilwell
```

dhartunian: ah thanks for actually reading the docs!

PR #169, opened 2021-01-05
db-console: alter base path with HOCs

koorosh:
Current change is an attempt to resolve the issue when route paths have to be altered with some base path provided from outside. All paths inside of `db-console` suppose that base path is `/` and in case client application has defined another base path for Statements page than these internal paths inside of statements page won't know about this and will be broken.

This change suggests following:
- `react-router` and navigation isn't used directly inside of pages - `history.push` and `history.replace` methods are wrapped inside of `useNavigation` hook which is a single point for calling this methods and it also has an access to `basePath` which can be defined with `BasePathContext`. It allows us to avoid usage of contexts directly in components and abstract logic in hook (and as HOC for class components).
- `useNavigation` hook should be used for functional components
- `withNavigation` HOC should be used for class components (provides `navigate` prop)

Pros:
- changing browsers history can be considered as a side effect and has to be called outside of component.
- entire logic of handling navigation and altering base path is defined in a single place

Cons:
- It is not possible to disallow calling `history.push/replace` methods from component and might break usage of base paths;

### DIFF
--- a/packages/admin-ui-components/src/index.ts
+++ b/packages/admin-ui-components/src/index.ts
@@ -23,5 +23,6 @@ export * from "./transactionsPage";
 export * from "./text";
 export * from "./tooltip";
 export * from "./tooltip2";
+export * from "./routeNavigation";
 export { util, api };
 export * from "./sessions";
--- a/packages/admin-ui-components/src/routeNavigation/basePathContext.ts
+++ b/packages/admin-ui-components/src/routeNavigation/basePathContext.ts
@@ -1,0 +1,3 @@
+import React from "react";
+
+export const BasePathContext = React.createContext("");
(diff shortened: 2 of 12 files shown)


dhartunian:
@laurenbarker would love your thoughts on this approach to avoid the nested router issue

koorosh:
Here is an example of required changes for client app which integrates Statements page for instance.
Client app has to wrap Statements page with `BasePathContext` to provide base path as a value so internal routing logic could rely on this value.
https://github.com/cockroachdb/cockroach/pull/58639/files#diff-9676030474fc4bd23dad78e71f392d79603cf3143964d844fba35f88bba66cc7L140-R154

PR #170, opened 2021-01-07
admin-ui-components: sessions page copy changes

nkodali:
This change updates the sessions page to use
'duration' instead of 'age' and 'transaction'
instead of 'txn'. This aims to make the meaning
of these fields less ambiguous.

Release note (ui change): On the sessions page in db console, any 'age'
label has been replaced with 'duration' and 'txn' has been replaced with
'transaction'. This was only a label change, with the actual metrics
remaining unchanged.

Relates to cockroachdb/cockroach#57047.


dhartunian (approved):
LGTM, just adjust the commit title length before merging :)

PR #174, opened 2021-01-15
db-console: optimize Webpack build process with production mode

koorosh:
Initially, webpack config was defined with hard coded "development"
mode that isn't sufficient for production.
Now, webpack's "mode" option is provided as an argument in package.json
to set production mode for builds.
Production mode provides a set of optimizations to minify bundle


dhartunian (approved):
Wonderful! Thank you.

PR #176, opened 2021-01-15
cluster-ui: optimize bundle size

koorosh:
There is five dependencies that mostly impacted the size of bundle: `crdb-protobuf-client`, `moment.js`, `lodash`, `highlight.js`, and `antd`.
- Moment.js by default imported all locales;
- highlight.js imported syntax configs for different languages (only SQL syntax is used)
- antd imported many unused components and styles.

Current changes handle these three libs (except `lodash`) with limiting imports to required functionalities.
It is easier to review this PR per commit with more detailed explanation in commit messages.

Current changes reduce the size of bundle from ~20Mb to ~14Mb

PS: `lodash` imports aren't optimized here because they require code changes (eliminate `_.chain` usage)
PS: `crdb-protobuf-client` size is ~6Mb from total size (50%) not minified. Require optimizations during package build.

- [x] tabulation for highlight SQL syntax is broken

### DIFF
--- a/packages/cluster-ui/babel.config.js
+++ b/packages/cluster-ui/babel.config.js
@@ -17,6 +17,7 @@ const presets = [
 const plugins = [
   "@babel/proposal-class-properties",
   "@babel/proposal-object-rest-spread",
+  ["import", { "libraryName": "antd", "style": "css" }]
 ];
 
 const env = {
--- a/packages/cluster-ui/package.json
+++ b/packages/cluster-ui/package.json
@@ -83,6 +83,7 @@
     "astroturf": "^0.10.2",
     "babel-jest": "^26.2.2",
     "babel-loader": "^8.0.6",
+    "babel-plugin-import": "^1.13.3",
     "babel-preset-react-app": "^9.1.2",
     "chai": "^4.2.0",
     "classnames": "^2.2.6",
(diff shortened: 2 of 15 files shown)


packages/admin-ui-components/babel.config.js
```diff
@@ -17,6 +17,7 @@ const presets = [
 const plugins = [
   "@babel/proposal-class-properties",
   "@babel/proposal-object-rest-spread",
+  ["import", { "libraryName": "antd", "style": "css" }]
```
koorosh, line 20: This configuration will require Antd components from `ant/lin/<component>` with relevant styles, instead of loading entire package and styles.

packages/admin-ui-components/src/tooltip/tooltip.tsx
```diff
@@ -1,5 +1,6 @@
 import React from "react";
-import AntTooltip, { AbstractTooltipProps } from "antd/lib/tooltip";
+import { Tooltip as AntTooltip } from "antd";
```
koorosh, line 2: All `antd` modules have to be imported as:
```import { Component } from "antd";``` not from `"antd/lib/<component>"`
because `babel-import-plugin` expects `antd` as libraryName (see babel.config.js file)

File: packages/admin-ui-components/webpack.config.js [hunk 3]
```diff
@@ -108,6 +109,7 @@ module.exports = {
       color: "cyan",
       profile: true,
     }),
+    new MomentLocalesPlugin(),
```
koorosh, line 112: By default `MomentLocalesPlugin` loads only `us` locale.

dhartunian (approved):
LGTM. Thanks for taking the time to do this!

PR #177, opened 2021-01-21
cluster-ui: nodes and liveness redux functionality

koorosh:
These change extends redux functionality with
two more reducers: nodes and node liveness statuses.
Reducers and sagas for these slices are implemented
in the same way as statements redux functionality,
with only one exception:
- nodes saga performs transformation for received
data to replicate the same logic as were before.


dhartunian (commented):
This all LGTM apart from the node accumulateMetrics application in the store. I'm sure we'll uncover bugs when testing in CC but for now we should ship it and see.

dhartunian (approved):
Feel free to merge once you move out the nodes accumulate call.

koorosh:
Sorry for merging this PR before applying last changes and having unnecessary revert PR.
This PR is reverted back because the last requested change was included and a new PR (https://github.com/cockroachdb/ui/pull/186) includes
all changes.

PR #178, opened 2021-01-22
cluster-ui: update statements page

asubiotto:
This commit adds columns for execution stats that we want to display for 21.1.
Namely, this adds rows/bytes read, max memory, and network bytes sent. The
rows affected column has been removed.
The commit also reorders all column-related code to follow the graphical order.

Here is what that looks like with some example queries run against tpch:
image

Closes https://github.com/cockroachdb/cockroach/issues/59294

Please let me know which tests to update and how. I have run into some issues with "accessing property mean of undefined" which is most likely due to unset proto fields. Is that something I need to handle in the protojs package in cockroach? If not, I'd appreciate if someone could point me to how to write tests for this and other cases.

I'm explicitly not adding code to handle cases when the stats haven't been collected. I leave this for the future if we decide to keep the sampling cluster setting. I focused on just getting the stats out there using preexisting components.

#### Checklist
- [x] Decide whether to remove the TXN Type column. The design doesn't have it, but want to double check cc @awoods187 @Annebirzin 

cc @awoods187 @cockroachdb/sql-execution

### DIFF
--- a/packages/cluster-ui/src/barCharts/barCharts.tsx
+++ b/packages/cluster-ui/src/barCharts/barCharts.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 import * as protos from "@cockroachlabs/crdb-protobuf-client";
 import { stdDevLong } from "src/util/appStats";
-import { Duration } from "src/util/format";
+import { Duration, Bytes } from "src/util/format";
 import classNames from "classnames/bind";
 import styles from "./barCharts.module.scss";
 import { bar, formatTwoPlaces, longToInt, approximify } from "./utils";
@@ -16,16 +16,12 @@ const countBars = [
   ),
 ];
 
-const retryBars = [
-  bar(
-    "count-retry",
-    (d: StatementStatistics) =>
-      longToInt(d.stats.count) - longToInt(d.stats.first_attempt_count),
-  ),
+const rowsReadBars = [
+  bar("rows-read", (d: StatementStatistics) => d.stats.rows_read.mean),
 ];
 
-const rowsBars = [
-  bar("rows", (d: StatementStatistics) => d.stats.num_rows.mean),
+const bytesReadBars = [
+  bar("bytes-read", (d: StatementStatistics) => d.stats.bytes_read.mean),
 ];
 
 const latencyBars = [
@@ -38,22 +34,56 @@ const latencyBars = [
   ),
 ];
 
+const maxMemUsageBars = [
+  bar("max-mem-usage", (d: StatementStatistics) => d.stats.max_mem_usage?.mean),
+];
+
+const networkBytesBars = [
+  bar("network-bytes", (d: StatementStatistics) => d.stats.bytes_sent_over_network?.mean),
+]
+
+const retryBars = [
+  bar(
+    "count-retry",
+    (d: StatementStatistics) =>
+      longToInt(d.stats.count) - longToInt(d.stats.first_attempt_count),
+  ),
+];
+
+const rowsReadStdDev = bar(cx("rows-read-dev"), (d: StatementStatistics) =>
+  stdDevLong(d.stats.rows_read, d.stats.exec_stat_collection_count),
+);
+const bytesReadStdDev = bar(cx("rows-read-dev"), (d: StatementStatistics) =>
+  stdDevLong(d.stats.bytes_read, d.stats.exec_stat_collection_count),
+);
 const latencyStdDev = bar(
   cx("bar-chart__overall-dev"),
   (d: StatementStatistics) => stdDevLong(d.stats.service_lat, d.stats.count),
 );
-const rowsStdDev = bar(cx("rows-dev"), (d: StatementStatistics) =>
-  stdDevLong(d.stats.num_rows, d.stats.count),
+const maxMemUsageStdDev = bar(
+  cx("max-mem-usage-dev"),
+  (d: StatementStatistics) =>
+    stdDevLong(d.stats.max_mem_usage, d.stats.exec_stat_collection_count),
+);
+const networkBytesStdDev = bar(
+  cx("network-bytes-dev"),
+  (d: StatementStatistics) =>
+    stdDevLong(d.stats.bytes_sent_over_network, d.stats.exec_stat_collection_count),
 );
 
 export const countBarChart = barChartFactory("grey", countBars, approximify);
-export const retryBarChart = barChartFactory("red", retryBars, approximify);
-export const rowsBarChart = barChartFactory(
+export const rowsReadBarChart = barChartFactory(
   "grey",
-  rowsBars,
+  rowsReadBars,
   approximify,
-  rowsStdDev,
+  rowsReadStdDev,
   formatTwoPlaces,
+);
+export const bytesReadBarChart = barChartFactory(
+  "grey",
+  bytesReadBars,
+  Bytes,
+  bytesReadStdDev,
 );
 export const latencyBarChart = barChartFactory(
   "grey",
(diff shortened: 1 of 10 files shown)


asubiotto:
Also, a note that I don't think this PR builds yet. To do so, I need to merge the protobuf updates I have in #59132

Annebirzin:
@asubiotto I believe we are planning to remove the 'TXN type' column because we now have the transactions page (but we would still add txn type as a filter and include it on the statement detail page) @awoods187 can you confirm?

awoods187:
@Annebirzin that matches my understanding.

@asubiotto this is exciting!

dhartunian:
@asubiotto I published a beta version of the CRDB client from this PR: https://github.com/cockroachdb/cockroach/pull/59132 which you can use to get this to pass build.

If any of the endpoints could return a missing value, you can guard against that using patterns like:
```
const maxMemUsageBars = [
  bar("max-mem-usage", (d: StatementStatistics) => d.stats.max_mem_usage?.mean),
];

const networkBytesBars = [
  bar(
    "network-bytes",
    (d: StatementStatistics) => d.stats.bytes_sent_over_network?.mean,
  ),
];
```
With the `?` on the field that may be `undefined`. Documented here: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining.

I'll look through code in more detail as well.

dhartunian:
@asubiotto use `0.0.5-beta.0` as the `crdb-protobuf-client` version in `cluster-ui/package.json`

packages/cluster-ui/src/statementsTable/statementsTable.tsx
```diff
@@ -34,33 +37,43 @@ function makeCommonColumns(
...
+      root: cx("statements-table__col-bytes-read--bar-chart"),
+      label: cx("statements-table__col-bytes-read--bar-chart__label"),
     },
   });
   const latencyBar = latencyBarChart(statements, {
     classes: {
       root: cx("statements-table__col-latency--bar-chart"),
     },
   });
+  const maxMemUsageBar = maxMemUsageBarChart(statements, {
```
dhartunian, line 57: Could we make the new column layout configurable via `makeCommonColumns` and a prop on `StatementsPage`? This component is now shipped to CockroachCloud so needs to remain backwards compatible with 20.2 at the very least.

dhartunian (commented):
Regarding testing, I think making sure that the Storybook pages for StatementsPage and StatementDetails page components render properly is a good place to start. You can adjust the fixtures in packages/cluster-ui/src/statementsPage/statementsPage.fixture.ts for example to make sure the pages have the input data they need for the new fields.

packages/cluster-ui/src/statementsTable/statementsTable.tsx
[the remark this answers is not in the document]
asubiotto: Do you mind showing me how to do this? I'm on board with the idea but not sure what you mean by "prop on `StatementsPage`. OOC, what does backward compatibility mean in this case? Allowing the user to toggle to the previous view?

asubiotto:
Added the fields to the storybook fixture and change the protobuf version to the beta you pushed (thanks). I also had a question about legend formatters. I haven't passed any into the bars I created, is that fine?

packages/cluster-ui/src/statementsTable/statementsTable.tsx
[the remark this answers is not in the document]
dhartunian: I was about to leave a comment to not worry about this actually. Our team is having a discussion about how to support multiple CRDB versions anyway and we'll make sure to address this issue ourselves.

packages/cluster-ui/src/barCharts/barCharts.tsx
```diff
@@ -38,26 +34,71 @@ const latencyBars = [
...
-  rowsStdDev,
+  rowsReadStdDev,
   formatTwoPlaces,
 );
+export const bytesReadBarChart = barChartFactory(
+  "grey",
+  bytesReadBars,
+  Bytes,
+  bytesReadStdDev,
+  // TODO(asubiotto): What's a legend formatter, do we need it?
```
dhartunian, line 83: @asubiotto Legend formatter can convert the numeric stat values to strings in the pop up legend using a *different* method than the "primary" formatter (`Bytes`) here for instance. I think in your case you want them to be consistent so you can omit the argument.

Here's an example where I added lots of numbers after the decimal point and just used `toString` as the legend formatter, while the primary formatter remained `approximify`.
Screen Shot 2021-01-26 at 1 26 08 PM

dhartunian:
I think what we'll do is once your CRDB change is merged, we'll publish a new proto version that's non-beta, then we can merge this, and you can bump the `cluster-ui` dependency version in `pkg/ui`.

asubiotto:
> I think what we'll do is once your CRDB change is merged, we'll publish a new proto version that's non-beta, then we can merge this, and you can bump the cluster-ui dependency version in pkg/ui.

Sounds good.

I fixed the build issue which was caused by my deleting the rows affected title which is still used by the transactions page (for now). Added the .scss back in. Related to this, is there a way to get a tool to yell at you if you're referencing a css class that doesn't exist?

packages/cluster-ui/src/statementsTable/statementsTable.module.scss
```diff
@@ -24,7 +24,7 @@
   }
```
asubiotto, line 24: Do I need to change anything here?

asubiotto:
Looks like there are some test issues which I believe https://github.com/cockroachdb/cockroach/pull/59132 fixes. I'll wait for that to merge.

Also, I see that I'm getting some lint errors and would like to run eslint --fix on save. Have you by any chance set that up in Goland? I tried using an automatic configuration which didn't work due to not being able to find an eslint package. I then tried to link it to both `eslint` and `typescript-eslint` in the parent `node_modules` folder but it complains about the absence of an `lib/options` folder.

dhartunian:
> > I think what we'll do is once your CRDB change is merged, we'll publish a new proto version that's non-beta, then we can merge this, and you can bump the cluster-ui dependency version in pkg/ui.
> 
> Sounds good.
> 
> I fixed the build issue which was caused by my deleting the rows affected title which is still used by the transactions page (for now). Added the .scss back in. Related to this, is there a way to get a tool to yell at you if you're referencing a css class that doesn't exist?

I don't think such a feature exists. The bar chart code in particular is a bit ugly so it's easier to make mistakes in there. Maybe @nathanstilwell has thoughts.

dhartunian:
@asubiotto I have this setup for eslint in Goland and it works for this project. You should only have to run `eslint` and it takes care of using `prettier` for formatting and parsing typescript.

Screen Shot 2021-01-27 at 1 18 23 PM

packages/cluster-ui/src/statementsTable/statementsTable.module.scss
```diff
@@ -24,7 +24,7 @@
...
 }
 
-.statements-table__col-count, .statements-table__col-retries, .statements-table__col-rows {
+.statements-table__col-count, .statements-table__col-rows, .statements-table__col-rows-read, .statements-table__col-bytes-read , .statements-table__col-max-mem-usage, .statements-tabe__col-network-bytes, .statements-table__col-retries {
```
dhartunian: there's a `tabe` -> `table` spelling error here. I will push a small commit to your branch that cleans this up. It's quite annoying to deal with and we should only need to use a single selector here.

packages/cluster-ui/src/statementsTable/statementsTable.module.scss
dhartunian: Nope, these are overrides for specific column widths in case you need them.

dhartunian (commented):
I pushed a small update with CSS edits that hopefully help, and I ran `yarn lint:fix` in a second commit.

asubiotto:
Thanks for the edits. Re eslint, I did try that but for some reason it doesn't work (regardless of wheter I'm using `ui` or `cluster-ui` project root) and asks me to specify a path to the `eslint` package. This does work in `cockroachdb/cockroach`. Will keep on fiddling with it.

asubiotto:
New version that points to v0.0.5 protobuf changes and fixes CI errors. I think this should be good to go! PTAL

dhartunian (approved):
LGTM.

Once you merge this, an automated build will create and publish the next version of the `cluster-ui` package. You can then bump the version in CRDB to get the latest version of the statements page. Latest version will be pushed as a git commit with a version number to the repo, or can be seen in npm: https://www.npmjs.com/package/@cockroachlabs/cluster-ui

asubiotto:
Thanks for your guidance!

PR #179, opened 2021-01-25
# cluster-ui // Truncating long queries with CSS

nathanstilwell:
# Truncating long queries

In this PR I'm truncating long queries in several places where they are displayed in tabular form through DB Console. There are a couple of CSS techniques that come into play (see commit messages and / or code for details) and there was not uniform place to make a unifying change. 

**Statements**
Screen Shot 2021-01-28 at 14 08 34

**Statements Application Filter**
Screen Shot 2021-01-28 at 14 07 30

**Transactions**
Screen Shot 2021-01-28 at 14 09 16

**Sessions**
Screen Shot 2021-01-28 at 14 10 09

### DIFF
--- a/packages/cluster-ui/src/statementsTable/statementsTable.module.scss
+++ b/packages/cluster-ui/src/statementsTable/statementsTable.module.scss
@@ -1,4 +1,4 @@
-@import 'src/core/index.module';
+@import "src/core/index.module";
 
 .statements-table__col-time {
   white-space: nowrap;
@@ -18,13 +18,16 @@
   }
 }
 
-.statements-table__col-rows, .statements-table__col-latency {
+.statements-table__col-rows,
+.statements-table__col-latency {
   &--bar-chart {
     min-width: 150px;
   }
 }
 
-.statements-table__col-count, .statements-table__col-retries, .statements-table__col-rows {
+.statements-table__col-count,
+.statements-table__col-retries,
+.statements-table__col-rows {
   &--bar-chart {
     margin-left: 0;
 
@@ -42,5 +45,9 @@
   width: 400px;
   div {
     font-size: $font-size--small;
+    display: -webkit-box;
+    -webkit-line-clamp: 2;
+    -webkit-box-orient: vertical;
+    overflow: hidden;
   }
 }
--- a/packages/cluster-ui/src/statementsTable/statementsTable.module.scss
+++ b/packages/cluster-ui/src/statementsTable/statementsTable.module.scss
@@ -45,9 +45,6 @@
   width: 400px;
   div {
     font-size: $font-size--small;
-    display: -webkit-box;
-    -webkit-line-clamp: 2;
-    -webkit-box-orient: vertical;
-    overflow: hidden;
+    @include line-clamp(2);
   }
 }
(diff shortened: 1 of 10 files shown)


dhartunian (approved):
TIL about `-webkit-line-clamp`. LGTM

dhartunian (changes requested):
Actually, could we add this to the Transactions page as well? I assume we don't use the same stylesheet on that page.

Annebirzin:
Reviewed this with Nathan over video, but noting that another place where we truncate long queries is on the sessions page and the jobs page. (The jobs page seems to already have truncation in place, but not sure about the sessions page).

Screen Shot 2020-12-10 at 4 18 42 PM

nathanstilwell:
@Annebirzin it dawned on me later, but I don't think the Sessions page has be extracted to cluster-ui yet. So that change needs to be another Jira issue so I can put a PR against CRDB.

dhartunian:
@nathanstilwell sessions page was merged into `cluster-ui` here: https://github.com/cockroachdb/ui/pull/164

nathanstilwell:
🤦🏼 Thanks @dhartunian. My bad @Annebirzin

dhartunian:
@nathanstilwell this PR contains a bunch of unrelated commits, can you rebase on `master`?

packages/cluster-ui/src/statementsTable/statementsTable.module.scss
```diff
@@ -1,4 +1,4 @@
-@import 'src/core/index.module';
+@import "src/core/index.module";
```
dhartunian, line 1: is there a linter we can add for this?

dhartunian (approved):
realized I could just take a look at the individual commits, might still want to rebase before merging 👍🏻

koorosh (commented):
@nathanstilwell, looks great! What do you think about refactoring this truncation styles into **mixing** which accepts the number of rows to truncate and define variables to keep the number of rows in tooltips and table rows?

koorosh:
@nathanstilwell , some unwanted commits get into this PR, it might be easier to remove them with interactive rebasing to avoid merge conflicts.

nathanstilwell:
@koorosh I was thinking about a mixin to consolidate the styles. I think eventually a component could also work, but I don't have a clear picture of that in mind yet. So a mixin feels like the right solution for now.

I also need to rebase to clear out some of these commits. Thanks for the call out.

packages/cluster-ui/src/statementsTable/statementsTable.module.scss
nathanstilwell: I'm using [Prettier](https://prettier.io/) which I thought we had already included in [@cockroachlabs/eslint-config](https://github.com/cockroachdb/ui/blob/master/packages/eslint-config/index.js). It seems like the lint task that runs in the npm script doesn't consider css files, so that's something we might change. I think I made the assumption that everyone uses linter integrated into their IDEs, but perhaps we should move toward making linting errors break the build. 

The issue there is that there are currently hundreds of linting errors in the `cluster-ui` and DB Console code.

koorosh (approved):
LGTM 🚀

packages/cluster-ui/src/statementsTable/statementsTable.module.scss
dhartunian: 👍🏻 didn't realize linting errors didn't break the build. I can look at that.

PR #180, opened 2021-01-26
cluster-ui: fix TS config for storybooks

koorosh:
Excluded .stories.tsx files from `tsconfig` caused issues that
ts rules weren't applied to storybooks and caused errors during
development (due to default ts rules were applied instead of
project specific rules from tsconfig.json file.
Exclusion of `.stories.tsx` files was intended to fix typings
generation for production build.

Final build has to include only sources from `./src` dir (this
dir is defined as a root dir in `tsconfig`), but
storybooks referenced to ./.storybook dir (which is outside of
`src` dir) and caused incorrect directory structure in `dist` dir.

With this change, all storybook custom decorators which is imported
within source code - moved inside `src` directory. This
structural change, provides proper output directory structure and
apply ts rules for all files in source directory.

As an additional change, storybook's backgroundColor decorator
is removed and instead, storybook configuration is provided with
default background color.


dhartunian:
LGTM, thanks @koorosh

PR #181, opened 2021-01-26
cluster-ui: provide ordered list of diagnostics to statements page

koorosh:
Due to missed type definition for returned type for
`selectDiagnosticsReportsPerStatement` selector, it caused
a regression issue when instead of returning the list of all
diagnostics reports per statement, this selector returned only last report
(according to previous and outdated requirements).

Now, return type is explicitly defined for `selectDiagnosticsReportsPerStatement`
 selector and it returns aggregated statement stats with all diagnostics reports
 per statement.

We need all reports on Statements page to show dropdown list with current state
for recently requested diagnostics and links for previously generated reports.

Before:
Screen Shot 2021-01-26 at 4 30 36 PM

After:
Screen Shot 2021-01-26 at 4 28 44 PM

### DIFF
--- a/packages/cluster-ui/src/statementsPage/statementsPage.selectors.ts
+++ b/packages/cluster-ui/src/statementsPage/statementsPage.selectors.ts
@@ -15,7 +15,8 @@ import { RouteComponentProps } from "react-router-dom";
 
 import { AppState } from "src/store";
 import { StatementsState } from "../store/statements";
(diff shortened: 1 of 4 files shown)


dhartunian (approved):

dhartunian:
Good reminder to eventually get to the point where we can turn on return type enforcement

PR #186, opened 2021-01-28
cluster-ui: nodes and liveness redux functionality

koorosh:
These change extends redux functionality with
two more reducers: nodes and node liveness statuses.
Reducers and sagas for these slices are implemented
in the same way as statements redux functionality.


koorosh:
This is exact same PR as https://github.com/cockroachdb/ui/pull/177 with one extra change requested - to move out node accumulators logic from sagas to selectors.

PR #193, opened 2021-02-04
cluster-ui: update transactions page with execution stats

asubiotto:
This PR has https://github.com/cockroachdb/cockroach/pull/59761 as a dependency. I'll just wait for that to go in before updating with the new protobuf version number.

This PR, similarly to #178, changes the transactions page to what we've designed for 21.1 (@Annebirzin). This is the result:

image

#### Checklist

cc @awoods187 

Closes https://github.com/cockroachdb/cockroach/issues/59295

### DIFF
--- a/packages/cluster-ui/src/transactionsPage/utils.ts
+++ b/packages/cluster-ui/src/transactionsPage/utils.ts
@@ -9,6 +9,7 @@ import { aggregateNumericStats, FixLong } from "../util";
 type Statement = protos.cockroach.server.serverpb.StatementsResponse.ICollectedStatementStatistics;
 type TransactionStats = protos.cockroach.sql.ITransactionStatistics;
 type Transaction = protos.cockroach.server.serverpb.StatementsResponse.IExtendedCollectedTransactionStatistics;
+type ExecStats = protos.cockroach.sql.IExecStats;
 
 export const getTrxAppFilterOptions = (
   transactions: Transaction[],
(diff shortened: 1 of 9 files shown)


asubiotto:
I didn't see any tests to update

packages/cluster-ui/src/transactionsPage/utils.ts
```diff
@@ -124,12 +125,52 @@ const withFingerprint = function(
...
+  };
+}
+
 // addTransactionStats adds together two stat objects into one using their counts to compute a new
 // average for the numeric statistics. It's modeled after the similar `addStatementStats` function
 function addTransactionStats(
   a: TransactionStats,
   b: TransactionStats,
-): TransactionStats {
+): Required<TransactionStats> {
```
asubiotto, line 173: I added this in because I thought that incorrect addition of transaction stats was causing some null property accesses, but it was in fact due to something else. However, I kept it in because I thought it's hygienic. When is this used? I didn't update my statements page PR to update the similar `addStatementStats` (however, something similar exists in the main cockroach repo, which is not the case for transactions). What's the consequence of this? Should I update `addStatementStats`?

dhartunian: Yes I like making changes like this. I think making the same change to `addStatementStats` is worth it.

Because the Statements Page has been integrated into CC, it has its own redux layer in `cluster-ui` that was built to work nicely with CC's redux store. That redux layer is *not* used with DB Console so there is some code duplication there. I think what we'll end up doing in the near term is moving the redux layer directly into CC to clarify that the two redux configurations are separate and make peace with some code duplication.

So making that change to `addStatementStats` will protect CC's version of the Statements Page from null lookups.

dhartunian (approved):
LGTM. These changes are all relatively straightforward, let me know if you need a protobuf library published, otherwise you're certainly free to push a V0.0.6 or whatever, yourself :)

PR #196, opened 2021-02-10
cluster-ui: add babel-runtime to support dynamic loading

koorosh:
Before, when components from `cluster-ui` packages were dynamically imported
into client code, it might fail to load with an error "regeneratorRuntime is
not defined" error that explained in following `babel` issue
https://github.com/babel/babel/issues/8829.
As a solution, `babel-runtime` and `@babel/plugin-transform-runtime` plugin are
 added in package.

### DIFF
--- a/packages/cluster-ui/babel.config.js
+++ b/packages/cluster-ui/babel.config.js
@@ -17,7 +17,9 @@ const presets = [
 const plugins = [
   "@babel/proposal-class-properties",
   "@babel/proposal-object-rest-spread",
-  ["import", { "libraryName": "antd", "style": "css" }]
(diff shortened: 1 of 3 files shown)


packages/cluster-ui/babel.config.js
```diff
@@ -17,7 +17,8 @@ const presets = [
 const plugins = [
   "@babel/proposal-class-properties",
   "@babel/proposal-object-rest-spread",
-  ["import", { "libraryName": "antd", "style": "css" }]
+  "@babel/plugin-transform-runtime",
```
dhartunian, line 20: Can you add a comment here to explain that this plugin was added to support dynamic loading into CC

dhartunian (approved):
LGTM

PR #197, opened 2021-02-10
cluster-ui: update transaction overview page with execution stats

asubiotto:
This PR implements the design of the transaction overview page. I modified some things from the original design and got to:

image

@Annebirzin / @awoods187 I'd like an answer specifically to:
- [x] I changed the "total" labels in the design to "mean" which I think is what we want, right? Total would be count of times this fingerprint was executed * mean latency, while mean latency is just mean latency and it doesn't matter how many times you ran the statement (the other would grow each time you run a statement, which I don't think makes sense)
- [x] I removed some data that seemed to be repeated in the design. PTAL and let me know if that's ok.

I also left a TODO to add temporary disk max usage, which requires a change on the backend.

Some CSS issues I need help with from @dhartunian or someone on the CC observability team:
- [x] How to display the new summary card on the same horizontal line as the sql statements. I tried to use `Row/Col` from `antd` but no change.
- [x] The idea is to make Mean Transaction Time larger than other stats (e.g. mean rows/bytes read) and reduce the size of Transaction resource usage. Should I use a summary card label for the latter with no value? The thing is that it seems that `h3/h4` sizes are ignored if used as labels.

Closes https://github.com/cockroachdb/cockroach/issues/59297

#### Checklist

### DIFF
--- a/packages/cluster-ui/src/transactionDetails/transactionDetails.tsx
+++ b/packages/cluster-ui/src/transactionDetails/transactionDetails.tsx
@@ -1,5 +1,7 @@
 import React from "react";
 import * as protos from "@cockroachlabs/crdb-protobuf-client";
+import classNames from "classnames/bind";
+
 import { makeStatementsColumns } from "../statementsTable";
 import {
   SortedTable,
@@ -17,15 +19,30 @@ import { SqlBox } from "../sql";
 import { aggregateStatements } from "../transactionsPage/utils";
 import Long from "long";
 import { Loading } from "../loading";
+import { SummaryCard } from "../summaryCard";
+import { Bytes, Duration, formatNumberForDisplay } from "src/util";
+
+import summaryCardStyles from "../summaryCard/summaryCard.module.scss";
+import transactionDetailsStyles from "./transactionDetails.modules.scss";
+import { Col, Row } from "antd";
+import { Text, Heading } from "@cockroachlabs/ui-components";
 
 const { containerClass } = tableClasses;
 
 type Statement = protos.cockroach.server.serverpb.StatementsResponse.ICollectedStatementStatistics;
+type TransactionStats = protos.cockroach.sql.ITransactionStatistics;
+
+const summaryCardStylesCx = classNames.bind(summaryCardStyles);
+const transactionDetailsStylesCx = classNames.bind(transactionDetailsStyles);
 
 interface TransactionDetailsProps {
   statements?: Statement[];
+  transactionStats?: TransactionStats;
   lastReset?: string | Date;
-  handleDetails: (statementIds: Long[] | null) => void;
+  handleDetails: (
+    statementIds: Long[] | null,
+    transactionStats: TransactionStats | null,
+  ) => void;
   error?: Error | null;
 }
 
@@ -61,12 +78,12 @@ export class TransactionDetails extends React.Component<
   };
 
   render() {
-    const { statements, handleDetails, error } = this.props;
+    const { statements, transactionStats, handleDetails, error } = this.props;
     return (
       <div>
         <section className={baseHeadingClasses.wrapper}>
           <Button
-            onClick={() => handleDetails(null)}
+            onClick={() => handleDetails(null, null)}
             type="unstyled-link"
             size="small"
             icon={BackIcon}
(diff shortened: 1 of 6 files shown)


asubiotto:
cc @nathanstilwell in case you also have input on the CSS/visual changes (last two unchecked boxes in the main comment)

Annebirzin:
@asubiotto 

- I will defer to @awoods187 to verify the 'mean' vs. 'total' time question
- regarding the redundant metrics in the design screenshot below, does `rows/bytes read over network`  `mean rows/bytes returned` and `network usage` all mean the same thing? cc @awoods187 I know we had discussed breaking these out at one point 
- Also, small nit, but wondering if we want to move `max memory` below both of the bytes read metrics 

Screen Shot 2021-02-12 at 5 00 06 PM

asubiotto:
Bytes over network means the same thing as network usage in bytes. Rows read over network is a proxy for bytes but might be useful to display separately. The `returned` metrics is not something we collect but could.

asubiotto:
Ack on moving max memory, will do it when responding to the first review.

asubiotto:
Friendly ping

awoods187:
I believe you are saying `Mean transaction time` = `Mean Latency` and `Total transaction time`=`Mean Latency` * `Total Number of Transactions`. I believe we care about showing `Mean transaction time`, `Mean Latency.` We can add a future enhancement (actually undeprecating a deprecated feature) for `Total transaction time` but we will need to call it something else because that name confuses people. Also, FWIW, I would show transaction time and latency and add mean to the tool tip. 

I vote for parallelism here: 
`Total rows/bytes read`
`Network rows/bytes read`
`Mean returned rows/bytes read`
`Max memory usage`
`Max scratch disk usage`

asubiotto:
I agree re mean.

> I would show transaction time and latency and add mean to the tool tip.

You mean change the label to remove "mean" and specify that it's mean txn time in the tooltip? I'd rather leave that for later since I want to do a pass later on to add tooltips etc...

packages/cluster-ui/src/transactionDetails/transactionDetails.tsx
```diff
@@ -95,6 +109,95 @@ export class TransactionDetails extends React.Component<
                     arrayItemName={"statements for this transaction"}
                     activeFilters={0}
                   />
+                  <SummaryCard>
+                    <div className={summaryCardStylesCx("summary--card__item")}>
+                      <h3
```
nathanstilwell, line 114: Here you could use the `<Heading />` component for `@cockroachlabs/ui-components`. It encompasses standard font treatments for the Design system. Usage could look something like, 

```jsx
import { Heading } from "@cockroachlabs/ui-components";

// .... 

<Heading type="h3" className={summaryCardStylesCx("summary--card__item--label")}>
  Mean transaction time
</h3>
```

Annebirzin:
👍 I updated the designs to reflect the correct wording for both the [transaction](https://www.figma.com/file/MoSPFkEyJfGipCmQvvLg1a/obsrv-query-performance?node-id=1763%3A25658) and [statement](https://www.figma.com/file/MoSPFkEyJfGipCmQvvLg1a/obsrv-query-performance?node-id=1888%3A68852) detail pages

Screen Shot 2021-02-17 at 11 47 06 AM

Screen Shot 2021-02-17 at 11 52 09 AM

packages/cluster-ui/src/transactionDetails/transactionDetails.tsx
asubiotto: Thanks! Didn't realize I could see that in figma.

asubiotto:
@Annebirzin one last thing. Can you confirm that we want to use the word "total" in "Total rows/bytes read" here? In fact we will be showing the mean number of rows read:
image

Thanks @koorosh and @nathanstilwell for the CSS help. I have included @koorosh's changes and this is the result:
image

This is ready for a review.

awoods187:
I think mean makes sense over total given the prior conversation

Annebirzin:
yep agree, "mean" makes sense

dhartunian (approved):
Don't think it's necessary for this PR but in the future, try and avoid importing styles from one component into another. Ideally the `summaryCardStyles` are only used by `SummaryCard` and related components. You could create companion components to `SummaryCard` like `SummaryCardHeading`, `SummaryCardDivider` that will give you building blocks for the more complex card.

I can see in this case the pattern is the same as in StatementDetails so we can keep it as is and refactor later.

asubiotto:
Thanks! Will keep that in mind.

PR #200, opened 2021-02-11
cluster-ui: multiple component versions

koorosh:
Components for different versions of CockroachDb are incompatible
and the latest version of `cluster-ui` package doesn't allow to
use components with previous (older) versions of CockroachDb.

This change splits down statements and statement details pages
into two versions where components for previous (20.2) version
have suffix `_20_2` and components without prefix - are targeted
to the latest CRDB version.

Maintaining several versions are needed in CC where cluster can
be spun up with arbitrary CRDB version and it has to conditionally
load components depending on cluster version.

This approach has obvious **disadvantages**:
- code duplication
- multiple versions of components don't inherit common logic
- version changes are tracked in code instead of VCS

Component duplicates have very small changes - most of the time it's
a one line change.

### DIFF
--- a/packages/cluster-ui/src/statementDetails/statementDetails.tsx
+++ b/packages/cluster-ui/src/statementDetails/statementDetails.tsx
@@ -37,6 +37,7 @@ import {
 import {
   AggregateStatistics,
   makeNodesColumns,
+  makeNodesColumns_20_2,
   StatementsSortedTable,
 } from "src/statementsTable";
 import { DiagnosticsView } from "./diagnostics/diagnosticsView";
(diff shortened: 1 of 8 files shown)


packages/cluster-ui/src/statementDetails/statementDetails.tsx
```diff
@@ -850,3 +851,595 @@ export class StatementDetails extends React.Component<
...
+                      i
+                    </div>
+                  </div>
+                </Tooltip>
+              </div>
+            </h2>
+            <StatementsSortedTable
+              className={cx("statements-table")}
+              data={statsByNode}
+              columns={makeNodesColumns_20_2(statsByNode, this.props.nodeNames)}
```
koorosh, line 1435: `...makeNodesColumns_20_2...`
this is the only difference between `StatementDetails` and `StatementDetails_20_2` components

koorosh:
Is not actual anymore. Closed

PR #207, opened 2021-02-11
Delete Group 392.png

Annebirzin:
# docs // Delete asset


Annebirzin:
testing deleting an asset

PR #213, opened 2021-02-17
cluster-ui: add full table or index scan filter to statements page

barryhe2000:
Add checkbox to filter statements by full table or index scan.


dhartunian (commented):
The implementation looks good so far but we should have it conform to some visuals prior to merging. Does this need to get incorporated into a dropdown?

awoods187:
Hi all, what's the status here? We'd love to get this into 21.1 and show it to customers.

barryhe2000:
The filter checkbox has been done for a while so the functionality is there, but matching Anne's design with the filter dropdown and keeping all of that functionality is what's taking a while...

barryhe2000:
Screen Shot 2021-03-01 at 3 22 10 PM

barryhe2000:
The filter dropdown is still missing some of its features so after finishing that integration the checkbox could just sit inside the dropdown and we'd get the same Figma design as seen in last week's weekly meeting, but right now it's a bit of a mess since there's still a few bugs I'm dealing with when trying to export out the filter dropdown from transactions page into statements

awoods187:
This is great progress. Should we merge the filter as is and then come back and adjust the ux? I'd like to test the functionality. Also, we can engage with our friends in observability if we need some extra help on the UX.

barryhe2000:
@dhartunian do you have any thoughts on merging in its current state? I know we already finished filtering last week but the UX side isn't quite done yet

barryhe2000:
Screen Shot 2021-03-02 at 4 15 02 PM
Screen Shot 2021-03-02 at 4 15 13 PM

Statements page with checkbox full scan filter

barryhe2000:
Update: plans are to first merge the checkbox in and look into the filter dropdown as a later issue. For now the ui state is as shown above, with the filter by full scans functionality working as intended.

packages/cluster-ui/src/statementsPage/statementsPage.tsx
```diff
@@ -265,6 +273,20 @@ export class StatementsPage extends React.Component<
...
+          <PageConfigItem>
+            <div>
+              <input
+                type="checkbox"
+                id="full-table-scan-toggle"
+                defaultChecked={this.state.fullScan}
+                onChange={this.fullScanChange}
+              />
+              <label htmlFor="full-table-scan-toggle">
+                &nbsp;&nbsp;Only show statements that contain queries with full
```
nathanstilwell, line 285: This is a _React thing_, but if you want to add literal spaces in HTML you can do something like 

```suggestion
                {"  "} Only show statements that contain queries with full
```

instead of using `&nbsp;`

nathanstilwell (approved):
I had one nit, but this LGTM.

barryhe2000:
TFTR! Merging.

PR #214, opened 2021-02-17
ui-components: 'reasonable' spelling was corrected

nanduu04:
'Reasonable' spelling was corrected on the tooltip storybook

### DIFF
--- a/packages/storybook-ui-components/stories/Tooltip.stories.tsx
+++ b/packages/storybook-ui-components/stories/Tooltip.stories.tsx
@@ -12,7 +12,7 @@ export default {
 export const Example = () => {
   const text = (
     <span>
-      reasonble tooltip text,
+      reasonable tooltip text,
       <br />
       some more text
     </span>


packages/storybook-ui-components/stories/Avatar.stories.tsx
```diff
@@ -1,47 +1,47 @@
-import React, { FunctionComponent, ReactElement } from "react";
-import { withKnobs, text, boolean, select } from "@storybook/addon-knobs";
-import { action } from "@storybook/addon-actions";
-import { Avatar, AvatarProps, Icon } from "@cockroachlabs/ui-components";
-import { StoryContainer, StoryDescription } from "../layout";
+import React, { FunctionComponent, ReactElement } from 'react';
```
dhartunian, line 1: Why does your formatter change all the quotes in this PR?

dhartunian (changes requested):
It's difficult to tell what you've changed when the diff contains so many formatting modifications. Can you push an updated commit without them?

packages/storybook-ui-components/stories/Avatar.stories.tsx
nanduu04: prettier had it change it automatically to single quotes, updated the changes.

dhartunian: Gotcha. We use `eslint` to apply `prettier` along with other linting rules. Make sure your formatter is configured that way. If you just "run prettier" on the code you won't get the correct result unfortunately.

packages/storybook-ui-components/stories/Tooltip.stories.tsx
```diff
@@ -12,7 +12,7 @@ export default {
 export const Example = () => {
   const text = (
     <span>
-      reasonble tooltip text,
+      reasonbale tooltip text,
```
dhartunian, line 15: is this meant to fix the spelling error, because it still needs adjustment

packages/storybook-ui-components/stories/Tooltip.stories.tsx
```diff
@@ -160,12 +160,12 @@ export const Demo = () => (
             "bottom-start",
             "bottom-end",
           ],
-          "bottom",
+          "bottom"
```
dhartunian, line 163: Trailing commas should also be handled by eslint. Make sure your editor is configured to use it.

packages/ui-components/src/Tooltip/Tooltip.module.scss
```diff
@@ -90,17 +90,30 @@ $tooltip-arrow-half-width: 5px;
 // Default style
 // --------------------------------------------------
 .default {
-  @include crl-tooltip-style($color-core-neutral-6, $color-core-neutral-6, $color-base-white);
+  @include crl-tooltip-style(
+    $color-core-neutral-6,
+    $color-core-neutral-6,
+    $color-base-white
+  );
+  z-index: 5;
```
dhartunian, line 98: If you're going to fix this using a `z-index` then we should have some agreement on what `z-index` values we expect to use in the app. Since this field is difficult to use correctly, people put all sorts of numbers into it to "make things work". Can you explain how you got to the number 5?

dhartunian (changes requested):

packages/ui-components/src/Tooltip/Tooltip.module.scss
nathanstilwell: @dhartunian I think the `z-index: 5` is a value that is already used in Tooltips for the arrow elements. But I am also suspicious of the use of z-index here. [Stacking Contexts](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) are notoriously tricky things to deal with in CSS. It is a common misconception that `z-index` is a global or shared value among elements in the DOM, which increases confusion. I haven't read the code yet, but I have to imagine that [`react-popper`](https://popper.js.org/react-popper/v2/) already has some consideration to control the stacking context around the target element.

For the life of me I can't recreate this issue on `master` in Storybook. Are we sure this is still and issue?

packages/storybook-ui-components/stories/Tooltip.stories.tsx
nanduu04: Done

packages/storybook-ui-components/stories/Tooltip.stories.tsx
nanduu04: Done

packages/ui-components/src/Tooltip/Tooltip.module.scss
nanduu04: Issue closed for now, reverted the changes

packages/storybook-ui-components/stories/Avatar.stories.tsx
nanduu04: Done

packages/ui-components/src/Tooltip/Tooltip.module.scss
```diff
@@ -90,17 +90,29 @@ $tooltip-arrow-half-width: 5px;
...
 //
 // Light style
 // --------------------------------------------------
 .light {
-  @include crl-tooltip-style($color-base-white, $color-core-neutral-2, $color-font-3);
+  @include crl-tooltip-style(
+    $color-base-white,
+    $color-core-neutral-2,
+    $color-font-3
+  );
```
abarganier, line 117: nit: since we're no longer touching this file, let's revert the formatting changes (I know this was just prettier doing it's thing).

abarganier (changes requested):
Just one small nit re: css formatting changes, then good to go!

packages/ui-components/src/Tooltip/Tooltip.module.scss
nanduu04: Done

abarganier (approved):
Thanks Nandu! LGTM

dhartunian (approved):

PR #215, opened 2021-02-18
cluster-ui: sessions, terminate modal notifications

elkmaster:
provide separate notifications saga and expose its action, so consumer
app can build around it its own notification handlers. small cleanups.

### DIFF
--- a/packages/cluster-ui/src/sessions/sessionsDetailsConnected.stories.tsx
+++ b/packages/cluster-ui/src/sessions/sessionsDetailsConnected.stories.tsx
@@ -16,7 +16,7 @@ import {
   Store,
 } from "redux";
 import { SessionDetailsPageConnected } from "./sessionDetailsConnected";
-import { AppState, rootReducer, sagas } from "../store";
+import { AppState, sagas, rootReducer } from "../store";
 import { Route } from "react-router-dom";
 
 const TEST_ID = "165f9819-2feb-40f0-0000-000000000001";
--- a/packages/cluster-ui/src/sessions/terminateQueryModal.tsx
+++ b/packages/cluster-ui/src/sessions/terminateQueryModal.tsx
@@ -16,23 +16,15 @@ import React, {
 } from "react";
 import { Modal } from "../modal";
 import { Text } from "../text";
-
-// import {cockroach} from "src/js/protos";
-// import {trackTerminateQuery} from "src/util/analytics/trackTerminate";
-
-// import ICancelQueryRequest = cockroach.server.serverpb.ICancelQueryRequest;
-type ICancelQueryRequest = any;
-
+import { ICancelQueryRequest } from "src/store/terminateQuery";
 export interface TerminateQueryModalRef {
   showModalFor: (req: ICancelQueryRequest) => void;
 }
 
 interface TerminateQueryModalProps {
-  //cancel: (req: ICancelQueryRequest) => void;
-  cancel?: (req: ICancelQueryRequest) => void;
+  cancel: (payload: ICancelQueryRequest) => void;
 }
 
-// tslint:disable-next-line:variable-name
 const TerminateQueryModal = (
   props: TerminateQueryModalProps,
   ref: React.RefObject<TerminateQueryModalRef>,
@@ -43,7 +35,6 @@ const TerminateQueryModal = (
 
   const onOkHandler = useCallback(() => {
     cancel(req);
-    //trackTerminateQuery();
     setVisible(false);
   }, [req, cancel]);
 
--- a/packages/cluster-ui/src/sessions/terminateSessionModal.tsx
+++ b/packages/cluster-ui/src/sessions/terminateSessionModal.tsx
@@ -14,24 +14,18 @@ import React, {
   useImperativeHandle,
   useState,
 } from "react";
+import { ICancelSessionRequest } from "src/store/terminateQuery";
 import { Modal } from "../modal";
 import { Text } from "../text";
-
-//import {cockroach} from "src/js/protos";
-//import ICancelSessionRequest = cockroach.server.serverpb.ICancelSessionRequest;
-type ICancelSessionRequest = any;
-//import {trackTerminateSession} from "src/util/analytics/trackTerminate";
 
 export interface TerminateSessionModalRef {
   showModalFor: (req: ICancelSessionRequest) => void;
 }
 
 interface TerminateSessionModalProps {
-  //cancel: (req: ICancelSessionRequest) => void;
-  cancel?: (req: ICancelSessionRequest) => void;
+  cancel: (payload: ICancelSessionRequest) => void;
 }
 
-// tslint:disable-next-line:variable-name
 const TerminateSessionModal = (
   props: TerminateSessionModalProps,
   ref: React.RefObject<TerminateSessionModalRef>,
@@ -42,7 +36,6 @@ const TerminateSessionModal = (
 
   const onOkHandler = useCallback(() => {
     cancel(req);
-    //trackTerminateSession();
     setVisible(false);
   }, [req, cancel]);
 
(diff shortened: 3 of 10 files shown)


packages/cluster-ui/src/store/terminateQuery/terminateQuery.sagas.ts
```diff
@@ -11,49 +11,74 @@
...
+interface CancelSessionWithCallbacks {
+  req: ICancelSessionRequest;
+  onCancelSessionSuccess?: () => void;
+  onCancelSessionError?: () => void;
+}
+
+export type CancelSessionPayload = XOR<
+  CancelSessionWithCallbacks,
+  ICancelSessionRequest
+>;
```
dhartunian, line 34: I don't think this is an easily understood pattern and it's making already complex code much harder to understand. Why not use a standard Union type `|` and do a switch when you need to decide which one to use?

packages/cluster-ui/src/store/terminateQuery/terminateQuery.sagas.ts
```diff
@@ -11,49 +11,74 @@
...
+  const req = action.payload.req || action.payload;
+  const { onCancelSessionSuccess, onCancelSessionError } = action.payload;
+  const terminateSessionRequest = new CancelSessionRequest(req);
+
   try {
     yield call(terminateSession, terminateSessionRequest);
     yield put(terminateQueryActions.terminateSessionCompleted());
     yield put(sessionsActions.invalidated());
     yield put(sessionsActions.refresh());
+    onCancelSessionSuccess && onCancelSessionSuccess();
```
dhartunian, line 48: If we just have a single notification dispatch can we avoid having to do all the null checks?

dhartunian (changes requested):
@elkmaster could we have a simpler API that's just `sendNotification` that takes arguments instead? This introduces a lot of clutter to the interfaces and forces us to change it any time we want a new notification. Ideally the page could add a new notification and use the single `send...` prop to dispatch without that being a breaking change.

This is making me think that we should go ahead with moving the redux/saga code into CC so we can just directly add the dispatch for CC notifications in the redux code for these components. Otherwise we're complicating the interfaces of the components for no good benefit.

elkmaster:
@dhartunian,
updated pr to expose only one prop `sendNotification`.

packages/cluster-ui/src/store/terminateQuery/terminateQuery.sagas.ts
elkmaster: Not really, still need null check to support api for dbconsole env. Will be not necessary if/when we move redux layer to CC.

packages/cluster-ui/src/store/terminateQuery/terminateQuery.sagas.ts
elkmaster: This approach needed to support 2 very different types of payloads. Standard union will show error something like "type  Foo do not correspond to type Foo | Bar, Bar is missing". We will have ability to remove all this workarounds as soon as move redux layer to CC. But it is needed for now.

dhartunian (changes requested):
@elkmaster You are introducing an API for sending notifications. You need to document this API is the commit/PR message so people know how to use it.

packages/cluster-ui/src/store/terminateQuery/terminateQuery.sagas.ts
```diff
@@ -11,49 +11,90 @@
...
-export type ICancelQueryRequest = cockroach.server.serverpb.CancelQueryRequest;
+export type ICancelSessionRequest = cockroach.server.serverpb.ICancelSessionRequest;
+export type ICancelQueryRequest = cockroach.server.serverpb.ICancelQueryRequest;
+
+export enum NotificationType {
+  Success,
+  Error,
+}
+
+export type SendNotificationType = (
```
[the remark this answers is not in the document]
elkmaster: yes

packages/cluster-ui/src/store/terminateQuery/terminateQuery.sagas.ts
```diff
@@ -11,49 +11,90 @@
...
+  action: PayloadAction<CancelSessionPayload>,
 ) {
-  const terminateSessionRequest = new CancelSessionRequest(action.payload);
+  const { sendNotification } = action.payload;
+  const req = action.payload.req || action.payload;
+  const terminateSessionRequest = new CancelSessionRequest(req);
+
   try {
-    yield call(terminateSession, terminateSessionRequest);
+    //yield call(terminateSession, terminateSessionRequest);
```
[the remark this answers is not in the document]
elkmaster: typo

dhartunian (changes requested):
I know I promised to approve this but I'm still not satisfied with where this PR stands. 

My hesitation stems from the fact that adding notifications currently *doubles* the length of the `terminateQuery.sagas.ts`. Can this be justified? It's a lot of extra code for something that is unrelated to the core of what the terminate query action should be doing.

I think we can solve this by simplifying the payload type that we dispatch. For now, I don't understand the value of having two different types of objects dispatched via `cancel` in the `TerminateSessionModal`. Instead of the XOR structure, I'd prefer always sending the wrapped version with an optional `sendNotification` field. It captures equivalent set of states while simplifying the call sites and reducing the amount of code in the saga.

packages/cluster-ui/src/sessions/sessionsDetailsConnected.stories.tsx
```diff
@@ -40,7 +40,13 @@ const store: Store<AppState> = createStore(
   ),
 );
```
elkmaster, line 42: here is an example of usage on consumer app side

packages/cluster-ui/src/sessions/terminateQueryModal.tsx
```diff
@@ -16,23 +16,18 @@ import React, {
 } from "react";
 import { Modal } from "../modal";
 import { Text } from "../text";
+import { ICancelQueryRequest } from "src/store/terminateQuery";
```
dhartunian, line 19: Remove commented out code from this file.

packages/cluster-ui/src/sessions/terminateSessionModal.tsx
```diff
@@ -14,24 +14,20 @@ import React, {
   useImperativeHandle,
   useState,
 } from "react";
+import { ICancelSessionRequest } from "src/store/terminateQuery";
```
dhartunian, line 17: Remove commented out code from this file

packages/cluster-ui/src/sessions/sessionsDetailsConnected.stories.tsx
```diff
@@ -40,7 +40,13 @@ const store: Store<AppState> = createStore(
...
 
-sagaMiddleware.run(sagas);
+sagaMiddleware.run(
+  initClusterUISagas({
```
dhartunian, line 44: Why not pass the `SagasContext` as the second param to `sagaMiddleware` and have it be an argument to `sagas` function after the cache invalidation period? That seems like a more straightforward use of the API

packages/cluster-ui/src/store/sagas.ts
```diff
@@ -16,6 +16,14 @@ export function* sagas(cacheInvalidationPeriod?: number) {
...
+    fork(terminateSaga, this?.sendNotification),
   ]);
 }
+
+export type SagasContext = {
+  sendNotification: SendNotification;
+};
+
+export function initClusterUISagas(context: SagasContext) {
+  return sagas.bind(context);
```
dhartunian, line 28: See note in `sessionsDetailsConnected.stories.tsx` for alternative pattern to use here. I would like avoid calling `.bind`

dhartunian (changes requested):

packages/cluster-ui/src/sessions/sessionsDetailsConnected.stories.tsx
elkmaster: cache invalidation period is an optional param.  we can't have 2 optional params in a row

dhartunian: Gotcha. Can we just have a single configuration param with two optional fields? I don't want to introduce complex `bind` usage and have params come from two different places. It makes the code difficult to read.

dhartunian (approved):
I like this improvement that separates the notifications logic from the rest of the app.

PR #217, opened 2021-02-19
storybook: link ui-components dependencies

koorosh:
# storybook-ui-components // link ui-components dependencies

`storybook-ui-components` package always depends on the latest
version of `ui-components` package and during development Storybook
has to apply all local changes.
To achieve this, `storybook-ui-components` installs `ui-components`
package as a linked dependency so it always has latest version.

Before, it relied on "next" tag that didn't work as expected and
it was a reason when local installs failed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,9 @@
 # ui
 Published assets for Cockroach Labs user interfaces
 
-Commands for local setup:
-
-`npm run init` - for initial run, will install top level dependencies, package level ones, run lerna link
-`npm start` - run in root of the project to run dev servers for all sub packages.<br/>
+## Development setup
+- Install all dependencies with Yarn (it resolves dependencies for nested workspaces and links them if necessary)
+- Run `build` command for all packages to have generated assets in place (needed for `icons` and `design-tokens` workspaces)
+```shell
+yarn install && yarn lerna run build
+```
--- a/packages/storybook-ui-components/package.json
+++ b/packages/storybook-ui-components/package.json
@@ -20,7 +20,7 @@
   "homepage": "https://github.com/cockroachdb/ui#readme",
   "dependencies": {
     "@cockroachlabs/icons": "^0.4.0",
-    "@cockroachlabs/ui-components": "next"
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -1,7 +1,9 @@
 # ui
 Published assets for Cockroach Labs user interfaces
 
-Commands for local setup:
-
-`npm run init` - for initial run, will install top level dependencies, package level ones, run lerna link
-`npm start` - run in root of the project to run dev servers for all sub packages.<br/>
\ No newline at end of file
+## Development setup
```
elkmaster, line 4: please make corresponding changes in packake.json as init command seems to be outdated

koorosh: Done

elkmaster (approved):

dhartunian:
I wonder if this will fix the yarn publish problems as well 🤔

packages/storybook-ui-components/package.json
```diff
@@ -20,7 +20,7 @@
   "homepage": "https://github.com/cockroachdb/ui#readme",
   "dependencies": {
     "@cockroachlabs/icons": "^0.4.0",
-    "@cockroachlabs/ui-components": "next"
+    "@cockroachlabs/ui-components": "link:packages/ui-components"
```
nathanstilwell, line 23: I like the look of this, but what happens during deploy? It just uses what's local to the repo? Which I guess is fine since it simply builds a static site.

nathanstilwell (approved):
Nice. I'm hoping this resolves some of the mysteries with deploy problems as well. LGTM. I had one simple question but shouldn't stop merging.

packages/storybook-ui-components/package.json
koorosh: > but what happens during deploy? 
good question! It will use local repo, but it might also require to build `ui-components` package as an extra step.

koorosh: @nathanstilwell , I've updated storybook publish configuration to build packages because local repository doesn't contain bundled files.

PR #223, opened 2021-02-24
cluster-ui: fix appstat utils

elkmaster:
additional null check to fix crashes in dbconsole envoirement

### DIFF
--- a/packages/cluster-ui/src/util/appStats.ts
+++ b/packages/cluster-ui/src/util/appStats.ts
@@ -12,7 +12,7 @@ export interface NumericStat {
 }
 
 export function variance(stat: NumericStat, count: number) {
-  return stat.squared_diffs / (count - 1);
+  return (stat?.squared_diffs || 0) / (count - 1);
 }
 
 export function stdDev(stat: NumericStat, count: number) {


asubiotto:
I assume this is related to my changes. I'd be interested in understanding why this would be happening. The proto fields are non-nullable on stats.

packages/cluster-ui/src/util/appStats.ts
```diff
@@ -20,7 +20,7 @@ export function stdDev(stat: NumericStat, count: number) {
 }
 
 export function stdDevLong(stat: NumericStat, count: number | Long) {
-  return stdDev(stat, Number(FixLong(count)));
+  return !!stat && stdDev(stat, Number(FixLong(count)));
```
koorosh, line 23: @elkmaster, it looks like `null` validation has to be handled inside of `variance` function. It will ensure that following functions (`variance`, `stdDev`, and `stdDevLong`) won't fail with this error.

```javascript
export function variance(stat: NumericStat, count: number) {
  return (stat?.squared_diffs || 0) / (count - 1);
}
```

koorosh (changes requested):

koorosh:
> I assume this is related to my changes. I'd be interested in understanding why this would be happening. The proto fields are non-nullable on stats.

that's really strange because generated js protobuf client defines all fields as optional and nullable 🤷‍♂️
Screen Shot 2021-02-24 at 10 35 24 PM

koorosh (approved):

asubiotto:
Huh, no idea why. Good to know though.

dhartunian:
@asubiotto I think this is how the TS generator makes the clients "backwards compatible" 😞

dhartunian (approved):

PR #226, opened 2021-02-26
cluster-ui: sample notification reducer 49d1ac2

dhartunian:
# PACKAGE // TITLE

#### Checklist

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/ui/226)


dhartunian:
@elkmaster had a thought about the notifications and made a little prototype...why don't we just force DB Console and CC to have custom reducers for notifications? The current path we're on seems to: pollute component props with notification-specific details and confuse sagas by introducing complex logic to account for dispatching notifications. Since the notifications are purely "decorative" and have nothing to do with the core functionality of the components, why not keep the logic for them completely outside and force the component user to write reducers using the exported actions that are available from our library?

Even in DB Console where we don't use the same redux layer, we should force the notifications code to live outside of the component.

See the second commit on this PR for my example.

PR #227, opened 2021-02-28
Refractor avatar unit tests to use react testing library

### LINKED ISSUE #81: [ui-components] refactor Avatar tests with RTL
- dependent on #49 
- refactor tests to implement RTL instead of Enzyme

tharun208:
# Refractor avatar unit tests to use react testing library

 - [x] Add `react testing library` library to package.json
 - [x] refractor `Avatar` component unit tests from to use `react testing library` and not  `enzyme`.

#### Checklist
- [x] I have written or updated the test for the changes I made

Fixes #81

### DIFF
--- a/packages/ui-components/src/Avatar/Avatar.test.tsx
+++ b/packages/ui-components/src/Avatar/Avatar.test.tsx
@@ -56,14 +56,12 @@ describe("Avatar", () => {
   describe("Content rendering", () => {
     it("renders with string passed as children", () => {
       render(<Avatar>foo bar</Avatar>);
-      const textContent = screen.getByTestId("avatar").textContent;
-      expect(textContent).toBe("foo bar");
+      expect(screen.getByTestId("avatar")).toHaveTextContent("foo bar");
     });
 
     it("renders with empty content", () => {
       render(<Avatar />);
-      const textContent = screen.getByTestId("avatar").textContent;
-      expect(textContent).toBe("");
+      expect(screen.getByTestId("avatar")).toHaveTextContent("");
     });
 
     it("renders with svg passed as children", () => {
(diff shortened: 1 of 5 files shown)


tharun208:
@asubiotto, can I get a review on this.

asubiotto:
Hi @tharun208, I'm not the best person to review this. cc @dhartunian

dhartunian:
Hi @tharun208 thanks for making these changes and submitting the PRs.

Your PRs are in the review pipeline so give us a few days to get to them. We're in the middle of wrapping up a release cycle for the DB so our response time on reviews will be a bit slow. If we haven't gotten to it by next Monday, please feel free to ping me again :) Appreciate your patience.

packages/ui-components/src/Avatar/Avatar.test.tsx
```diff
@@ -1,75 +1,91 @@
 import React from "react";
-import { shallow } from "enzyme";
+import { render, fireEvent, screen } from "@testing-library/react";
 
 import Avatar, { AvatarIntent, AvatarSize } from "./Avatar";
```
nathanstilwell, line 4: I don't want to pollute the component with a data attribute, but I agree that having one on the component is convenient for testing. So maybe we could wrap the component from the lib to add a data attribute in this file for testing purposes. I'm imagining something like,

```jsx
import UiAvatar, { AvatarIntent, AvatarSize, AvatarProps } from "./Avatar";

const Avatar: React.FC<AvatarProps> = ({ children, ...props }) => (
  <UiAvatar data-testid="avatar" {...props}>
    {children}
  </UiAvatar>
);
```

nathanstilwell (commented):
The tests look great, but I would rather not add a data attribute to the component. I left a comment for a possible solution that would minimize the refactoring.

packages/ui-components/src/Avatar/Avatar.test.tsx
tharun208: updated the tests @nathanstilwell

PR #231, opened 2021-03-01
cluster-ui: aggregate stmn data on txn details page

elkmaster:
cluster-ui: aggregate stmn data on txn details page …

**before**: multiple copies of same statements was show on
transaction details page. copies have same id but different
node id.
Screenshot from 2021-03-04 11-32-00

**after:** aggregate function used, duplicates eliminated.added
basic unit test for aggregate method.
Screenshot from 2021-03-04 11-30-25

CRDB-2366


dhartunian (changes requested):
Nice work, looks simple enough.

Can you either post screenshots of before/after for clarity OR add a test case for ensuring that the data on the Transactions + Transaction Details page matches up for the number of statements computed from the fixture. It's just not obvious "why" this works from looking at the diff.

elkmaster:
@dhartunian, done both

PR #234, opened 2021-03-02
cluster-ui: reset redux state action

koorosh:
# cluster-ui // reset redux state action

Initially, it wasn't possible to reset redux state at some point
during application life cycle.
Switching between multiple clusters that should keep their own
state in redux store requires to have ability to re-initialize
an empty redux state.


dhartunian (approved):
👍🏻 LGTM

nathanstilwell (approved):
👍🏼

PR #240, opened 2021-03-08
cluster-ui: update tooltips and add "no samples" tooltip on statements/transaction pages

asubiotto:
This PR updates the tooltip text according to this figma design: https://www.figma.com/file/MoSPFkEyJfGipCmQvvLg1a/obsrv-query-performance?node-id=1834%3A2895

Additionally, the second commit adds an option to the bar chart to display a message of "no samples" accompanied with a tooltip when a sampled execution stat has not been sampled yet. This is more user friendly as we would previously show a zero value. This is a screenshot for the second commit:

image

Closes https://github.com/cockroachdb/cockroach/issues/61346
Closes https://github.com/cockroachdb/cockroach/issues/61607

### DIFF
--- a/packages/cluster-ui/src/barCharts/barChartFactory.tsx
+++ b/packages/cluster-ui/src/barCharts/barChartFactory.tsx
@@ -3,6 +3,7 @@ import { extent as d3Extent } from "d3-array";
 import _ from "lodash";
 import React from "react";
 import { Tooltip } from "src/tooltip";
+import { Tooltip2 } from "src/tooltip2";
 import classNames from "classnames/bind";
 import styles from "./barCharts.module.scss";
 import { NumericStatLegend } from "./numericStatLegend";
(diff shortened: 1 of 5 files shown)


dhartunian (approved):
:lgtm: nice ux change!

> Reviewed 2 of 2 files at r1, 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/ui/240#-:-MVMeA5vAokO7hocduq5:b-iuaqe0)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Annebirzin and @nathanstilwell)

asubiotto:
TFTR! @Annebirzin. Are we sure we want the "black tooltip" background for the no samples? This is different than the tooltips we currently have on bar charts:
image

Annebirzin:
@asubiotto right now the pattern we follow is that we use the dark tooltip for text-only, and the light tooltip for the bar chart visualizations so I'd leave as is for now.

Question about the bar chart tooltips: I believe the metrics used to appear next to the bars like in the screenshot below - were those bars removed? (that was part of the reasoning for using the light background tooltips for bar tooltips) 
Screen Shot 2021-03-09 at 2 26 54 PM

asubiotto:
Gotcha. It seems like the bars were indeed removed. This seems like a separate investigation so a new issue sounds good.

Annebirzin:
@asubiotto sounds good, added an issue here https://github.com/cockroachdb/ui/issues/244 thanks!

PR #249, opened 2021-03-15
cluster-ui: ui customization parameters

koorosh:
Before, it wasn't necessary to customize UI components depending
on specific runtime conditions.
But now, it has to be possible to allow some client application
code to define which components have to be hidden on UI or shown.

In this particular change, "Statement diagnostics" link has to be
hidden because an entire Statement Diagnostics page might not be
available in client's code.

UI customization is implemented within redux store as separate
slice where all required customization params can be defined
with their default values.

To change some params, client's application has to dispatch
uiConfig's "update" action with partially filled configuration that
includes only params which have to be changed.

Screen Shot 2021-03-15 at 6 22 07 PM

### DIFF
--- a/packages/cluster-ui/src/statementDetails/diagnostics/diagnosticsView.tsx
+++ b/packages/cluster-ui/src/statementDetails/diagnostics/diagnosticsView.tsx
@@ -26,6 +26,7 @@ type IStatementDiagnosticsReport = cockroach.server.serverpb.IStatementDiagnosti
 export interface DiagnosticsViewStateProps {
   hasData: boolean;
   diagnosticsReports: cockroach.server.serverpb.IStatementDiagnosticsReport[];
+  showDiagnosticsViewLink?: boolean;
 }
 
 export interface DiagnosticsViewDispatchProps {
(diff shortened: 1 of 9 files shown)


nathanstilwell (commented):
The code looks good and makes sense to me, but I'm wondering if there is somewhere we could document  when the `showDiagnosticsViewLink` prop should be set to `false`. This option is going to be utilized by the application, right? Will it be obvious as to when this option should be used?

koorosh:
> The code looks good and makes sense to me, but I'm wondering if there is somewhere we could document when the `showDiagnosticsViewLink` prop should be set to `false`. This option is going to be utilized by the application, right? Will it be obvious as to when this option should be used?

Good question! That's definitely has to be somewhere documented! Would it be okay to put it in `uiConfig.reducer.ts` file or it's better to add it in the main `readme`?

nathanstilwell:
I think a comment in the reducer is good for now. I'm going to give this a thumbs up, but wonder if there's a way to programmatically determine this _hiding / showing_ based on data. Then the user of the component wouldn't have to worry about it (other than an override for a special circumstance).

nathanstilwell (approved):
LGTM

PR #250, opened 2021-03-16
cluster-ui:  fix statements page styles

koorosh:
Different pages use different implementations of table component
with their very own styles or override default styles with custom.
It caused styles mismatching.

Current change doesn't try to unify all components/styles but
locally fixes styles and somehow reuses it. It'll allow to
backport changes for older versions with less stress.

Before & After screens:
Screen Shot 2021-03-16 at 2 10 42 PM
----
Screen Shot 2021-03-16 at 2 09 51 PM
----
Screen Shot 2021-03-16 at 12 15 37 PM

### DIFF
--- a/packages/cluster-ui/src/sortedtable/tableHead/tableHead.module.scss
+++ b/packages/cluster-ui/src/sortedtable/tableHead/tableHead.module.scss
@@ -1,38 +1,38 @@
 @import '../../core/index.module.scss';
 @import '../table.module.scss';
 
+@mixin table-header-text {
+  @include text--heading-6;
+  color: $colors--neutral-6;
+  font-weight: 600;
+  font-family: $font-family--semi-bold;
+  letter-spacing: 1.5px;
+}
+
 .head-wrapper {
   @include table-base;
 
-  .__cell--header {
+  .cell-header {
     border-right: 1px solid $colors--neutral-2;
     width: 250px;
   }
   &__row--header {
     border-bottom: 1px solid $colors--neutral-3;
-    height: 48px;
+    height: $line-height--large;
+    padding: $spacing-medium-small;
+    background-color: $colors--neutral-1;
     // Display the sort direction arrow in header cells.
     .sorted__cell {
-      padding: 12.55px 10px 11.55px;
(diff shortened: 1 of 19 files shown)


packages/cluster-ui/src/sortedtable/tableHead/tableHead.module.scss
```diff
@@ -1,31 +1,40 @@
...
+  font-weight: 600;
+  font-family: $font-family--semi-bold;
+  letter-spacing: 1.5px;
+}
+
 .head-wrapper {
   @include table-base;
 
-  &__cell--header {
+  .__cell--header {
```
nathanstilwell, line 15: I know it's weird that we still have BEM in CSS modules, but as we are updating these styles can we lose the underscores? It's nitpicky and certainly not required for this change, but something to keep in mind as we refactor our styles.

packages/cluster-ui/src/table/table.module.scss
```diff
@@ -1,26 +1,31 @@
 @import "src/core/index.module";
+@import "src/sortedtable/tableHead/tableHead.module";
 
 .crl-table-wrapper {
-  .ant-table {
+  :global(.ant-table) {
```
nathanstilwell, line 5: so the idea here is that we are allowing the application to define global styles?

packages/cluster-ui/src/sortedtable/tableHead/tableHead.module.scss
koorosh: Agree, renamed to "cell-header" 👍🏼

packages/cluster-ui/src/table/table.module.scss
koorosh: It just tells to treat ".ant-table" class as a global one because it's defined outside of this CSS module. 
`:global` annotation prevents CSS modules to change the name of this class so this selector can work with global Antd styles.

nathanstilwell: I see. I don't love the way this looks, but I don't love using the Ant Table, and it looks like it works. 😄

nathanstilwell (approved):
LGTM

PR #252, opened 2021-03-19
packages/cluster-ui: add % of all runtime to statements

maryliag:
Adding new column for `% of all runtime` into statements page

`% of runtime all statements with this fingerprint represent, compared to the cumulative runtime of all queries within the last hour or specified time interval.`

Issue: https://github.com/cockroachdb/cockroach/issues/61475 

This PR also fixes some ui on details page to align with the mockup (change of weights of fonts, spacing)

### DIFF
--- a/packages/cluster-ui/src/barCharts/barCharts.tsx
+++ b/packages/cluster-ui/src/barCharts/barCharts.tsx
@@ -1,11 +1,11 @@
-import React from "react";
 import * as protos from "@cockroachlabs/crdb-protobuf-client";
 import { stdDevLong } from "src/util";
-import { Duration, Bytes } from "src/util/format";
+import { Duration, Bytes, Percentage } from "src/util/format";
 import classNames from "classnames/bind";
 import styles from "./barCharts.module.scss";
 import { bar, formatTwoPlaces, longToInt, approximify } from "./utils";
-import { barChartFactory } from "./barChartFactory";
+import { barChartFactory, BarChartOptions } from "./barChartFactory";
+import { AggregateStatistics } from "src/statementsTable/statementsTable";
 
 type StatementStatistics = protos.cockroach.server.serverpb.StatementsResponse.ICollectedStatementStatistics;
 const cx = classNames.bind(styles);
@@ -127,3 +127,21 @@ export const networkBytesBarChart = barChartFactory(
 );
 
 export const retryBarChart = barChartFactory("red", retryBars, approximify);
+
+export function workloadPctBarChart(
+  statements: AggregateStatistics[],
+  defaultBarChartOptions: BarChartOptions<any>,
+  totalWorkload: number,
+) {
+  return barChartFactory(
+    "grey",
+    [
+      bar(
+        "pct-workload",
+        (d: StatementStatistics) =>
+          (d.stats.service_lat.mean * longToInt(d.stats.count)) / totalWorkload,
+      ),
+    ],
+    v => Percentage(v, 1, 1),
+  )(statements, defaultBarChartOptions);
+}
--- a/packages/cluster-ui/src/statementDetails/statementDetails.tsx
+++ b/packages/cluster-ui/src/statementDetails/statementDetails.tsx
@@ -20,6 +20,7 @@ import {
   stdDev,
   getMatchParamByName,
   formatNumberForDisplay,
+  calculateTotalWorkload,
 } from "src/util";
 import { Loading } from "src/loading";
 import { Button } from "src/button";
@@ -408,6 +409,7 @@ export class StatementDetails extends React.Component<
     );
 
     const statsByNode = this.props.statement.byNode;
+    const totalWorkload = calculateTotalWorkload(statsByNode);
     const logicalPlan =
       stats.sensitive_info && stats.sensitive_info.most_recent_plan_description;
     const duration = (v: number) => Duration(v * 1e9);
@@ -431,8 +433,8 @@ export class StatementDetails extends React.Component<
                 <Row>
                   <Col>
                     <div className={summaryCardStylesCx("summary--card__item")}>
-                      <Heading type="h5">Statement time</Heading>
-                      <Text>
+                      <Heading type="h5">Mean statement time</Heading>
+                      <Text type="body-strong">
                         {formatNumberForDisplay(
                           stats.service_lat.mean,
                           duration,
@@ -440,7 +442,7 @@ export class StatementDetails extends React.Component<
                       </Text>
                     </div>
                     <div className={summaryCardStylesCx("summary--card__item")}>
-                      <Text type="body-strong">Planning time</Text>
+                      <Text>Planning time</Text>
                       <Text>
                         {formatNumberForDisplay(stats.plan_lat.mean, duration)}
                       </Text>
@@ -449,7 +451,7 @@ export class StatementDetails extends React.Component<
                       className={summaryCardStylesCx("summary--card__divider")}
                     />
                     <div className={summaryCardStylesCx("summary--card__item")}>
-                      <Text type="body-strong">Execution time</Text>
+                      <Text>Execution time</Text>
                       <Text>
                         {formatNumberForDisplay(stats.run_lat.mean, duration)}
                       </Text>
@@ -461,11 +463,13 @@ export class StatementDetails extends React.Component<
                 </Row>
               </SummaryCard>
               <SummaryCard className={cx("summary-card")}>
-                <Heading type="h5">Resource usage</Heading>
                 <Row>
                   <Col>
                     <div className={summaryCardStylesCx("summary--card__item")}>
-                      <Text type="body-strong">Mean rows/bytes read</Text>
+                      <Heading type="h5">Resource usage</Heading>
+                    </div>
+                    <div className={summaryCardStylesCx("summary--card__item")}>
+                      <Text>Mean rows/bytes read</Text>
                       <Text>
                         {formatNumberForDisplay(
                           stats.rows_read.mean,
@@ -476,7 +480,7 @@ export class StatementDetails extends React.Component<
                       </Text>
                     </div>
                     <div className={summaryCardStylesCx("summary--card__item")}>
-                      <Text type="body-strong">Max memory usage</Text>
+                      <Text>Max memory usage</Text>
                       <Text>
                         {formatNumberForDisplay(
                           stats.exec_stats.max_mem_usage.mean,
@@ -485,7 +489,7 @@ export class StatementDetails extends React.Component<
                       </Text>
                     </div>
                     <div className={summaryCardStylesCx("summary--card__item")}>
-                      <Text type="body-strong">Network usage</Text>
+                      <Text>Network usage</Text>
                       <Text>
                         {formatNumberForDisplay(
                           stats.exec_stats.network_bytes.mean,
@@ -494,7 +498,7 @@ export class StatementDetails extends React.Component<
                       </Text>
                     </div>
                     <div className={summaryCardStylesCx("summary--card__item")}>
-                      <Text type="body-strong">Max scratch disk usage</Text>
+                      <Text>Max scratch disk usage</Text>
                       <Text>
                         {formatNumberForDisplay(
                           stats.exec_stats.max_disk_usage.mean,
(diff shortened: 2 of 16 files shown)


packages/cluster-ui/src/util/format.ts
```diff
@@ -101,11 +101,18 @@ export const BytesFitScale = (scale: string) => (bytes: number) => {
 /**
  * Percentage creates a string representation of a fraction as a percentage.
```
asubiotto, line 102: nit: add precision comment/example

packages/cluster-ui/src/statementDetails/statementDetails.tsx
```diff
@@ -431,16 +431,16 @@ export class StatementDetails extends React.Component<
                 <Row>
                   <Col>
                     <div className={summaryCardStylesCx("summary--card__item")}>
-                      <Heading type="h5">Statement time</Heading>
-                      <Text>
+                      <Heading type="h5">Mean statement time</Heading>
+                      <Text type="body-strong">
```
asubiotto, line 435: nit: don't worry about it now, but in general I would've put these style changes in a separate commit

packages/cluster-ui/src/barCharts/barCharts.tsx
```diff
@@ -63,6 +63,15 @@ const retryBars = [
   ),
 ];
 
+let total: number;
```
asubiotto, line 66: We should avoid making a global variable. Is there a way to pass `total` down to the bar chart as an argument?

packages/cluster-ui/src/statementsPage/statementsPage.tsx
```diff
@@ -227,6 +228,16 @@ export class StatementsPage extends React.Component<
       );
   };
 
+  calculateTotalWorkload = (statements: AggregateStatistics[]) => {
+    let totalWorkload = 0;
+    statements.forEach(
```
asubiotto, line 233: This is an O(n) operation on each page reload if I'm not mistaken. Is there a way to cache results and figure out when we need to recompute (i.e. statement state changes, or a filter has changed)?

I don't think this is a big deal but there might be something simple we can do. If not, it's fine to add a comment. Feel free to argue against me if you think it doesn't matter that much to recompute or if I'm wrong.

packages/cluster-ui/src/statementsPage/statementsPage.module.scss
```diff
@@ -53,6 +53,10 @@ h2.base-heading {
   width: 100%;
 }
 
+.table-area {
```
asubiotto, line 56: nit: this is what fixes the resize bug, right? I would've put this in a separate commit as well but don't worry about it now.

packages/cluster-ui/src/statementsTable/statementsTable.tsx
```diff
@@ -129,6 +143,7 @@ export interface AggregateStatistics {
   drawer?: boolean;
   firstCellBordered?: boolean;
   diagnosticsReports?: cockroach.server.serverpb.IStatementDiagnosticsReport[];
+  totalWorkload?: Long;
```
asubiotto, line 146: nit: I think some comments here on `totalWorkload` and elsewhere would be useful

asubiotto (changes requested):
Great work! Great to see this. On a macro level, I think this PR should've been split into three commits: one to add the % runtime stat, one for style changes, and one for the resize bug fix. It doesn't really matter now and this PR is not huge, but splitting changes into different logical commits is very useful for reviewers to understand which file changes correspond to which behavior changes.

One other thing: could you add screenshots of your changes?

packages/cluster-ui/src/statementsPage/statementsPage.module.scss
maryliag: that's correct, it was for the resizing. I'll keep in mind for future commits :)

packages/cluster-ui/src/barCharts/barCharts.tsx
maryliag: done

packages/cluster-ui/src/util/format.ts
maryliag: done

packages/cluster-ui/src/statementsTable/statementsTable.tsx
maryliag: done

maryliag:
Statements table before:
Screen Shot 2021-03-22 at 12 24 13 PM
After:
Screen Shot 2021-03-22 at 12 24 32 PM

Resizing issue before:
Screen Shot 2021-03-22 at 12 24 55 PM
After:
Screen Shot 2021-03-22 at 12 25 06 PM

Statements Detail page before:
Screen Shot 2021-03-22 at 12 25 28 PM
After:
Screen Shot 2021-03-22 at 12 25 36 PM

Transaction Details page before (it also had the same resize issue as the statements), also changed the back option to include the arrow and remove the `all` to be aligned with statements details page:
Screen Shot 2021-03-22 at 12 26 01 PM
After:
Screen Shot 2021-03-22 at 12 26 22 PM

packages/cluster-ui/src/statementsPage/statementsPage.tsx
maryliag: Initially I was going to create the value on the selector and it would only be updated when statements were updated (I would also need to recalculate on filter changes), but because we notice that crdb and ui are using different selectors, it would need to be updated on both, they need to be in sync, and so on, so the suggestion I got was to work only on the ui component and not touch on selectors, since they're considering merging those two.

When a filter is done, it needs to be recalculate anyway, the improvements is when the user refreshes the page, is that a common behaviour? users refreshing the page frequently? If they do, I can make the changes on both selectors, if that is not such a frequent case we can leave as is and I can create a issue to improve this on the future, when we have only one set of selectors

packages/cluster-ui/src/statementsPage/statementsPage.tsx
```diff
@@ -227,6 +228,16 @@ export class StatementsPage extends React.Component<
       );
   };
 
+  calculateTotalWorkload = (statements: AggregateStatistics[]) => {
+    let totalWorkload = 0;
```
nathanstilwell, line 232: You could drop this variable by using a `reduce` here instead of a `forEach`,

```jsx
  calculateTotalWorkload = (statements: AggregateStatistics[]) =>
    statements.reduce(
      (totalWorkload, stmt) =>
        (totalWorkload +=
          longToInt(stmt.stats.count) * stmt.stats.service_lat.mean),
      0,
    );

```

packages/cluster-ui/src/statementsPage/statementsPage.tsx
nathanstilwell: Since the `statements` in this case are derived from filtered data and none of this is in the Redux store, the only thing I can think to do is memoize this function somehow. I think this could work if you use a [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) with the statements data as the key and the result of the calculation as the value.

packages/cluster-ui/src/transactionDetails/transactionDetails.tsx
```diff
@@ -80,6 +82,16 @@ export class TransactionDetails extends React.Component<
     this.setState({ pagination: { ...pagination, current } });
   };
 
+  calculateTotalWorkload = (statements: Statement[]) => {
```
nathanstilwell, line 85: Is this the same function that was in `statementsPage.tsx`? If you extract it to a utility it might be more compelling to memoize it and probably write a test for it.

maryliag: the function itself does the same thing, but statements on this file is `Statement[]` and on the file `statementsPage.tsx` is a `AggregateStatistics[]`

maryliag:
Added a memoized total workload function and tests
feel free to review the PR again @asubiotto

packages/cluster-ui/src/statementsTable/statementsTable.tsx
```diff
@@ -143,6 +143,7 @@ export interface AggregateStatistics {
   drawer?: boolean;
   firstCellBordered?: boolean;
   diagnosticsReports?: cockroach.server.serverpb.IStatementDiagnosticsReport[];
+  // totalWorkload is the sum of service latency of all statements listed on the table
```
asubiotto, line 146: nit: add a period at the end of the comment.

packages/cluster-ui/src/util/memoize.ts
```diff
@@ -0,0 +1,48 @@
...
+import { AggregateStatistics } from "src/statementsTable";
+
+type Statement = protos.cockroach.server.serverpb.StatementsResponse.ICollectedStatementStatistics;
+type statementType = AggregateStatistics | Statement;
+type statementsType = Array<statementType>;
+
+const cache = new Map();
+
+/**
+ * a simple hash function to minimaze the key size on the memoized function's cache
```
asubiotto, line 12: nit: minimaze

packages/cluster-ui/src/util/memoizeTotalWorkload.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+import { aggStatFix } from "./memoizeTotalWorkload.fixture";
+
+describe("Calculating total workload with memoized function", () => {
+  it("calculating total workloaad with one statement", () => {
+    const result = memoizedTotalWorkload([aggStatFix]);
+    // Using approximately because float handling by javascript is imprecise
+    assert.approximately(result, 48.421019, 0.0000001);
+  });
+
+  it("calculating total workloaad with no statements", () => {
```
nathanstilwell, line 13: `workloaad` typo. 😄

packages/cluster-ui/src/util/memoize.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+/**
+ * Memoized function to calculate total workload of statements
+ * It will only recalculate if the list of statements change
+ * @param statements array of statements (AggregateStatistics or Statement)
+ * @returns the total workload of all statements
+ */
+export function memoizedTotalWorkload(statements: statementsType) {
+  const key = stringToHash(JSON.stringify(statements));
```
asubiotto, line 35: I wonder if calculating the statement hash now becomes even more expensive than just recalculating the service latency because we're stringifying the statements object every time we want to get the statement hash. I think we should either:
1) Calculate this hash when first populating this statements data, so we can use a precomputed hash here (if the statements don't change, the hash won't either).
2) Forget about this for now. We might be overcomplicating this.

I'm happy to do 2. Over optimization is the root of all evil.

asubiotto (changes requested):
Looking good! Just want to point out that the general workflow at CRL when making changes to existing commits is to amend and rebase existing commits rather than creating new ones to address comments. This makes it easier to review (at least in the main repo).

packages/cluster-ui/src/util/memoize.ts
maryliag: made the changes to use a non memoized function

nathanstilwell: Talking with @maryliag last night about this we were thinking through the same thing. It's the choice between either using the _stringified_ statements array as a key and risking the hit to memory or iterating over the array to get a unique value to cache the calculation all to save the cost of the calculation. I say we drop this and fix the performance problem when it arises. 

For posterity perhaps we can leave a nice descriptive comment about the performance and that there may be an opportunity to optimize in the future.

packages/cluster-ui/src/statementsTable/statementsTable.tsx
```diff
@@ -31,6 +32,7 @@ const longToInt = (d: number | Long) => Number(FixLong(d));
 
 function makeCommonColumns(
   statements: AggregateStatistics[],
+  totalWorkload?: number,
```
asubiotto, line 35: Is it correct to make this optional? When is it not specified?

The other problem with making it optional is that I think if it's not specified that's an almost guaranteed divide by 0 exception below, right?

maryliag: good point! Fixed

asubiotto (approved):
LGTM! :shipit:

PR #254, opened 2021-03-22
cluster-ui: fix npm build

elkmaster:
having ts code outside src is corrupting type generation on package
build. moving decorators in test-utils folder under src/.


packages/cluster-ui/src/test-utils/index.ts
```diff
@@ -1 +1,2 @@
 export * from "./testStoreProvider";
+export * from "./decorators";
```
koorosh, line 2: Trailing empty line is missed.

koorosh (approved):
LGTM, just a tiny request to add trailing line at EOF (comment below)

nathanstilwell (approved):
LGTM.

PR #255, opened 2021-03-23
backport 20.2: ui customization parameters

koorosh:
#### Backport for https://github.com/cockroachdb/ui/pull/249

Before, it wasn't necessary to customize UI components depending
on specific runtime conditions.
But now, it has to be possible to allow some client application
code to define which components have to be hidden on UI or shown.

In this particular change, "Statement diagnostics" link has to be
hidden because an entire Statement Diagnostics page might not be
available in client's code.

UI customization is implemented within redux store as separate
slice where all required customization params can be defined
with their default values.
To change some params, client's application has to dispatch
uiConfig's "update" action with partially filled configuration.


nathanstilwell (approved):
LGTM

PR #257, opened 2021-03-23
backport 20.2: fix statements page styles

koorosh:
#### Backport for https://github.com/cockroachdb/ui/pull/250

Different pages use different implementations of table component
with their very own styles or override default styles with custom.
It caused styles mismatching.

Current change doesn't try to unify all components/styles but
locally fixes styles and somehow reuses it. It'll allow to
backport changes for older versions with less stress.


nathanstilwell (approved):
LGTM

PR #258, opened 2021-03-24
cluster-ui: re-export uiConfiguration actions

koorosh:
cluster-ui package has to re-export uiConfig actions to allow external apps
to change the reducers state in type-safe way - with action creator and
provided types.


nathanstilwell (approved):
LGTM

PR #259, opened 2021-03-24
backport 20.2: reset redux state action

koorosh:
#### backport for https://github.com/cockroachdb/ui/pull/234

Initially, it wasn't possible to reset redux state at some point
during application life cycle.
Switching between multiple clusters that should keep their own
state in redux store requires to have ability to re-initialize
an empty redux state.

(cherry picked from commit 8e52a107eb3283c4ed60cca01499a866eeb717d4)

### DIFF
--- a/packages/cluster-ui/src/store/reducers.ts
+++ b/packages/cluster-ui/src/store/reducers.ts
@@ -1,4 +1,5 @@
-import { combineReducers } from "redux";
+import { combineReducers, createStore } from "redux";
+import { createAction, createReducer } from "@reduxjs/toolkit";
 import { StatementsState, reducer as statements } from "./statements";
 import { LocalStorageState, reducer as localStorage } from "./localStorage";
 import {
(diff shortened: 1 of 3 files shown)


packages/cluster-ui/src/store/reducers.ts
```diff
@@ -39,3 +41,16 @@ export const rootReducer = combineReducers<AdminUiState>({
...
+export const rootActions = {
+  resetState: createAction(`${DOMAIN_NAME}/RESET_STATE`),
+};
+
+/**
+ * rootReducer consolidates reducers slices and cases for handling global actions related to entire state.
+ **/
+export const rootReducer = createReducer(undefined, builder => {
+  builder
+    .addCase(rootActions.resetState, () => createStore(reducers).getState())
```
nathanstilwell, line 54: this is an interesting way to get the initial state. It makes total sense, but not what I expected! Love it.

nathanstilwell (approved):
LGTM

PR #253, opened 2021-03-22
cluster-ui: correct design for empty state in statements table

koorosh:
Statements table and statement diagnostics table display empty state
as embedded placeholder inside of table (when table header is visible).
According to new design, empty state placeholders should replace entire
table.
Current change changes placeholders in statements table and diagnostics
tables. Also some refactoring is done in Empty panel component to reuse
components from `ui-components` package and clean up hand made styles.

Figma design: https://www.figma.com/file/C85d0mrz8dkNs7Nw5mcTmf/obsrv-statements?node-id=1143%3A2455

Screen Shot 2021-03-22 at 12 47 32 PM

PR #261, opened 2021-03-26
cluster-ui: proper empty state design for statement diagnostics view

koorosh:
Several commits ago (62a6d372a55619625ae41f0b9dc4632a82ec3cee), empty state component for Statement diagnostics view has been
changed to old design and now it is rolled back.

Additional changes were made to reuse `ui-components` (`Heading`, `Text` and `Button` components).

This change fixes issue described https://github.com/cockroachdb/ui/pull/253#issuecomment-806903656

New designs:
Screen Shot 2021-03-26 at 9 31 05 PM

### DIFF
--- a/packages/cluster-ui/src/statementDetails/diagnostics/diagnosticsView.tsx
+++ b/packages/cluster-ui/src/statementDetails/diagnostics/diagnosticsView.tsx
@@ -1,24 +1,24 @@
-import React from "react";
+import React, { useCallback } from "react";
 import { Link } from "react-router-dom";
 import moment from "moment";
 import classnames from "classnames/bind";
 import { cockroach } from "@cockroachlabs/crdb-protobuf-client";
-import { Download } from "@cockroachlabs/icons";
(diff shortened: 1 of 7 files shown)


packages/cluster-ui/src/statementDetails/diagnostics/diagnosticsView.tsx
```diff
@@ -130,16 +158,11 @@ export class DiagnosticsView extends React.Component<
...
-                  <Button
-                    size="small"
-                    type="flat"
-                    iconPosition="left"
-                    icon={<Download />}
-                  >
-                    Bundle (.zip)
-                  </Button>
-                </a>
+                  <Download />
```
nathanstilwell, line 163: You could also use the `Icon` component from `ui-components`.

```jsx
import { Icon } from "@cockroachlabs/ui-compoents";

<Icon iconName="Download" />
```

This is only sugar on top to conform to standard sizes or hues which doesn't seem necessary here.

PR #262, opened 2021-03-29
cluster-ui: default values for statement details ui configuration

koorosh:
Recently Statement details page were extended with `uiConfig` prop that
includes flags to tweak visibility for particular UI elements.
`uiConfig` prop should not be required, and it should be okay to
skip it to render default UI.

Current change adds default value for `uiConfig` prop and makes this
field as optional for statementDetailsProps type.


maryliag:
LGTM

maryliag (approved):

PR #267, opened 2021-04-01
cluster-ui: analytics tracking events

koorosh:
Current change extends list of tracked events and refactors
analytics reducer to make it more generic.

Analytics reducer provides single action (`track`) that accepts
different event types - it allows extend this action to work with
new types and have single to watch in case client's app is interested
in all possible analytics events.

Event types include required `name` field that has to follow Segment's
naming conventions and the rest of optional fields can be considered as
Segments properties.

To avoid coupling of analytics events with components - all logic are
defined in Connected components. Connected components have been refactored
`mapDispatchToProps` object is defined as function to be able to dispatch
multiple events in single method (`onActivateStatementDiagnostics` method
dispatches one action to create report and next action to track analytics).

Following events tracked:
- activate statements diagnostics (from statements and statement details pages)
- download statement diagnostics archive
- click on statement row in statements table
- change App filter on Statements page
- search statement on statements page
- change tabs on statement details page
- soring by columns in tables (statements table, and statement details
tables: Diagnostics and Stats table).
- track when click `Back to statements` button on statement details page;

### DIFF
--- a/packages/cluster-ui/src/statementDetails/diagnostics/diagnosticsView.tsx
+++ b/packages/cluster-ui/src/statementDetails/diagnostics/diagnosticsView.tsx
@@ -32,6 +32,11 @@ export interface DiagnosticsViewDispatchProps {
   activate: (statementFingerprint: string) => void;
   dismissAlertMessage: () => void;
   onDownloadDiagnosticBundleClick?: (statementFingerprint: string) => void;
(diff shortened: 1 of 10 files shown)


elkmaster (approved):

nathanstilwell (approved):
LGTM

PR #251, opened 2021-03-19
cluster-ui: clean up tooltip components

elkmaster:
before: in `cluster-ui` package we used legacy versions of
tootips based on antd and with a bunch of code duplicates and
hacks.

after: used `ui-components` tooltip component, removed unneeded
code and styles. also on process of work find out that barchats
legend tooltips missing styles - added them.

use tooltip component from `ui-components` library instead local
versions based on antd. to achieve this `ui-componets/Tooltip`
also needed some updates:
 - provide new style for table title usage (it have same color
   scheme as `default` but different typographic style)
 - add ability to hover on tooltip for clickable elements inside
 - other minor style updates and fixes

as side effect of this work there is a fix of dev process, updates of 
storybook and webpack config was needed to run build
of `cluster-ui` with linked local `ui-components`.

PR #271, opened 2021-04-06
cluster-ui: add button to reset SQL stats on statement page and txn page

Azhng:
cluster-ui: add button to reset SQL stats on statement page and txn page

Addresses https://github.com/cockroachdb/cockroach/issues/33315

Before: 

Stmt page
old-stmt

Txn Page
old-txn

After:

Stmt Page:
new-stmt

Txn Page:
new-txn

Tooltip
new-tool-tip

### DIFF
--- a/packages/cluster-ui/src/tableStatistics/tableStatistics.module.scss
+++ b/packages/cluster-ui/src/tableStatistics/tableStatistics.module.scss
@@ -1,0 +1,8 @@
+.flex-display {
+  display: flex;
+}
+
+.tooltip-hover-area {
+  padding-top: 3px;
+  padding-right: 10px;
+}
--- a/packages/cluster-ui/src/tableStatistics/tableStatistics.tsx
+++ b/packages/cluster-ui/src/tableStatistics/tableStatistics.tsx
@@ -5,8 +5,15 @@ import { statisticsClasses } from "../transactionsPage/transactionsPageClasses";
 import { ISortedTablePagination } from "../sortedtable";
 import { Button } from "src/button";
 import { ResultsPerPageLabel } from "src/pagination";
+import { Tooltip } from "../tooltip";
+import statementStyles from "src/statementDetails/statementDetails.module.scss";
+import tableStatsStyles from "./tableStatistics.module.scss";
+import classNames from "classnames/bind";
+import { Icon } from "@cockroachlabs/ui-components";
 
 const { statistic, countTitle, lastCleared } = statisticsClasses;
+const cxStmt = classNames.bind(statementStyles);
+const cxStats = classNames.bind(tableStatsStyles);
 
 interface TableStatistics {
   pagination: ISortedTablePagination;
@@ -16,7 +23,11 @@ interface TableStatistics {
   activeFilters: number;
   search?: string;
   onClearFilters?: () => void;
+  resetSQLStats: () => void;
 }
+
+const toolTipText = `Statement history is cleared once an hour by default, which can be configured with the cluster setting
+ diagnostics.reporting.interval. Clicking ‘Clear SQL stats’ will reset SQL stats on the statements and transactions pages.`;
 
 const renderLastCleared = (lastReset: string | Date) => {
   return `Last cleared ${moment.utc(lastReset).format(DATE_FORMAT)}`;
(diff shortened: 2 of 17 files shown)


packages/cluster-ui/src/tableStatistics/tableStatistics.tsx
```diff
@@ -16,12 +22,30 @@ interface TableStatistics {
...
 }
 
+const toolTipText = `Statement history is cleared once an hour by default, which can be configured with the cluster setting
+ diagnostics.reporting.interval. Clicking ‘Clear SQL stats’ will reset SQL stats on the statements and transactions pages.`;
+
 const renderLastCleared = (lastReset: string | Date) => {
   return `Last cleared ${moment.utc(lastReset).format(DATE_FORMAT)}`;
 };
 
+const renderResetSQLStats = (resetSQLStats: () => void) => {
```
maryliag, line 35: for both render reset stats and tooltip there is no need to pass parameters (since they're both static) and they look very simple, so I don't think is worth creating a render function for each one (the last cleared render requires some "calculation", so it's better to leave it separated), but I would suggest add those two renders directly

Azhng: Done

packages/cluster-ui/src/tableStatistics/tableStatistics.tsx
```diff
@@ -5,8 +5,14 @@ import { statisticsClasses } from "../transactionsPage/transactionsPageClasses";
 import { ISortedTablePagination } from "../sortedtable";
 import { Button } from "src/button";
 import { ResultsPerPageLabel } from "src/pagination";
+import { Tooltip } from "../tooltip";
```
nathanstilwell, line 8: Just a head's up, we are consolidating tooltips to use the `Tooltip` from `@cockroachlabs/ui-components`. See #251.

@elkmaster, I was looking at this PR to see how replacing the `Tooltip` component might work and ran into two things,
1. We don't extend `React.HTMLAttributes<HTMLDivElement>` props in Tooltip (or any html attribute props) and so passing a `className` prop won't be applied.
2. I tried adding `Tooltip` here in the PR and got some `z-index` issues. Any ideas?
Screen Shot 2021-04-07 at 15 36 09

elkmaster: In my pr z-index issue is fixed. Regarding classname, in this specific case it's just "float: right". I am sure we can handle it without passing inside tooltip. Lets merge this pr first and I will handle this all on my side.

packages/cluster-ui/src/tableStatistics/tableStatistics.tsx
```diff
@@ -54,7 +65,18 @@ export const TableStatistics: React.FC<TableStatistics> = ({
       <h4 className={countTitle}>
         {activeFilters ? resultsCountAndClear : resultsPerPageLabel}
       </h4>
-      <h4 className={lastCleared}>{renderLastCleared(lastReset)}</h4>
+      <div>
```
nathanstilwell, line 68: Right now you are floating the children of this div. I think you could get the same effect by adding a "display:flex" to this container and not need to apply styles to the individual children.

nathanstilwell (changes requested):

packages/cluster-ui/src/tableStatistics/tableStatistics.module.scss
```diff
@@ -0,0 +1,8 @@
+.align-float {
+  float: left;
+}
+
+.tooltip-hover-area {
+  padding-top: 3px;
+  padding-right: 10px;
+}
```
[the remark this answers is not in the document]
Azhng: Done.

packages/cluster-ui/src/tableStatistics/tableStatistics.tsx
```diff
@@ -54,7 +65,18 @@ export const TableStatistics: React.FC<TableStatistics> = ({
...
+        <Tooltip text={toolTipText} className={cxStats("align-float")}>
+          <div className={cxStats("tooltip-hover-area")}>
+            <div className={cxStmt("numeric-stats-table__info-icon")}>i</div>
```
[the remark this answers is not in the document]
Azhng: Done.

packages/cluster-ui/src/tableStatistics/tableStatistics.tsx
Azhng: Done. :+1:

Azhng:
TFTR!

PR #272, opened 2021-04-07
cluster-ui: cancel active sagas when redux state is reset

koorosh:
Previous implementation of redux-sagas for fetching new data from server was
straightforward and didn't take into account cases such as resetting an entire
redux state, so it cause cases when request to some data is dispatched, then
redux state is reset, and after this sagas would handle returned data as
usual (which in fact causes inconsistent store state).

To avoid this cases, sagas have to be cancelled when redux state is reset
and all responses after this should be ignored. To achieve desired behavior,
most of the changes are done within custom `throttleWithReset` effect.
- now, it's behavior is changed to take and process only latest actions
instead of processing every incoming action. It guaranties that all previous
actions and forked tasks are cancelled.
- `throttleWithReset` effect doesn't buffer actions and ignores all except
current action.

Sagas, that use `throttleWithReset` effect provide one more cancellation action
pattern to watch: `rootActions.resetState`. This action is dispatched when redux
state is reset and initiates cancellation process for running saga's tasks.

To follow the same pattern, statement diagnostics sagas are refactored in the
same was as other sagas, to use `throttleWithReset` effect.

### DIFF
--- a/packages/cluster-ui/src/store/liveness/liveness.sagas.ts
+++ b/packages/cluster-ui/src/store/liveness/liveness.sagas.ts
@@ -3,6 +3,7 @@ import { getLiveness } from "src/api/livenessApi";
 import { actions } from "./liveness.reducer";
 
 import { CACHE_INVALIDATION_PERIOD, throttleWithReset } from "src/store/utils";
+import { rootActions } from "../reducers";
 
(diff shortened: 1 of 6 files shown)


nathanstilwell (approved):
LGTM

PR #275, opened 2021-04-08
cluster-ui: sessions style fixes

elkmaster:
updated css to fix corresponding style issues that was caused by
global css of managed service. disabled link for gateway node on
sessions details page.


packages/cluster-ui/src/sessions/sessionDetails.tsx
```diff
@@ -55,6 +56,7 @@ interface OwnProps {
   refreshNodesLiveness: () => void;
   cancelSession: (payload: ICancelSessionRequest) => void;
   cancelQuery: (payload: ICancelQueryRequest) => void;
+  uiConfig: UIConfigState["pages"]["sessionDetails"];
```
koorosh, line 59: `uiConfig` prop has to be considered as optional. Reusing this component in CRDB might require to set it.

packages/cluster-ui/src/sessions/sessionPage.module.scss
```diff
@@ -0,0 +1,16 @@
...
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+@import "../core/index.module.scss";
+
+.sessions-page{
+  all: initial;
+  font-family: $font-family--base;
+}
```
nathanstilwell, line 16: newline at the end of file please. 😄

nathanstilwell (commented):
Besides Andrii's comment, looks good to me.

koorosh (changes requested):
@elkmaster , I found following places that still have to be fixed:
- on sessions page, header size is large and has missed ".base-heading" class
- on session details page, text in table cells inherits global font size from `<span>` elements (CC). Should persist its own styles with font-size 14px.

packages/cluster-ui/src/sessions/sessionDetails.tsx
elkmaster: done

packages/cluster-ui/src/sessions/sessionPage.module.scss
elkmaster: done

elkmaster:
@koorosh, thanks for review. regarding last comment, are you sure that correct version of package used, not cached one? because I do not see mentioned issues:
Screenshot from 2021-04-19 15-04-47

koorosh (approved):
LGTM

PR #276, opened 2021-04-08
cluster-ui: sql type filter

maryliag:
Cluster-ui
- Update of `react-select` from version `1.2.1` to `4.3.0`  and `@types/react-select` from version `1.2.6` to `4.0.14` and changes to fix compatibility with the new version
- Creation of new multi select component
- New filter on Statements Page for SQL Type (DDL, DML, DCL and TCL)

Resolves: https://github.com/cockroachdb/cockroach/issues/49178

Screen Shot 2021-04-08 at 3 01 40 PM
Screen Shot 2021-04-08 at 3 01 47 PM
Screen Shot 2021-04-08 at 3 01 57 PM
Screen Shot 2021-04-08 at 3 03 28 PM

### DIFF
--- a/packages/cluster-ui/src/multiSelectCheckbox/multiSelectCheckbox.tsx
+++ b/packages/cluster-ui/src/multiSelectCheckbox/multiSelectCheckbox.tsx
@@ -1,0 +1,118 @@
+import React from "react";
+import Select, { components, OptionsType } from "react-select";
+import styles from "./multiSelectCheckbox.module.scss";
+import classNames from "classnames/bind";
+
+const cx = classNames.bind(styles);
+
+const defaultSelectProps = {
+  searchable: false,
+  clearable: false,
+};
+
+export interface SelectOption {
+  label: string;
+  value: string;
+  isSelected: boolean;
+}
+
+export interface MultiSelectCheckboxProps {
+  field: string;
+  options: SelectOption[];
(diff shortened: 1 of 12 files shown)


packages/cluster-ui/src/multiSelectCheckbox/multiSelectCheckbox.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+export interface MultiSelectCheckboxProps {
+  field: string;
+  options: SelectOption[];
+  parent: any;
+  placeholder: string;
+  value?: SelectOption[];
+}
+
+const CheckboxOption = (props: any) => {
```
nathanstilwell, line 27: I don't love how much we are using `any` here for these internal props of `react-select`. It's frustrating, because under the hood, they also use any. Could you perhaps add some comments about what these types represent if we have their type as `any`?

nathanstilwell (approved):
LGTM (with a request for more comments around the use of `any`)

packages/cluster-ui/src/multiSelectCheckbox/multiSelectCheckbox.tsx
maryliag: I added some comment, let me know if it's enough

PR #278, opened 2021-04-15
cluster-ui: sessions tracking events

elkmaster:
expose analytics events on sessions:
 - on table sorted
 - on session link clicked
 - on terminate session
 - on terminate statement
and sessions details page:
 - on back to sessions clicked
 - on terminate session
 - on terminate statement

---
This change is [Reviewable](https://reviewable.io/reviews/cockroachdb/ui/278)

### DIFF
--- a/packages/cluster-ui/src/sessions/sessionDetails.tsx
+++ b/packages/cluster-ui/src/sessions/sessionDetails.tsx
@@ -63,6 +63,10 @@ interface OwnProps {
   cancelSession: (payload: ICancelSessionRequest) => void;
   cancelQuery: (payload: ICancelQueryRequest) => void;
   uiConfig?: UIConfigState["pages"]["sessionDetails"];
+  onBackButtonClick?: () => void;
+  onTerminateSessionClick?: () => void;
+  onTerminateStatementClick?: () => void;
+  onStatementClick?: () => void;
 }
 
 export type SessionDetailsProps = OwnProps & RouteComponentProps;
@@ -112,7 +116,8 @@ export class SessionDetails extends React.Component<SessionDetailsProps, {}> {
   }
 
   backToSessionsPage = () => {
-    const { history, location } = this.props;
+    const { history, location, onBackButtonClick } = this.props;
+    onBackButtonClick && onBackButtonClick();
     history.push({
       ...location,
       pathname: "/sessions",
@@ -121,7 +126,13 @@ export class SessionDetails extends React.Component<SessionDetailsProps, {}> {
 
   render() {
     const sessionID = getMatchParamByName(this.props.match, sessionAttr);
-    const { sessionError, cancelSession, cancelQuery } = this.props;
+    const {
+      sessionError,
+      cancelSession,
+      cancelQuery,
+      onTerminateSessionClick,
+      onTerminateStatementClick,
+    } = this.props;
     const session = this.props.session?.session;
     const showActionButtons = !!session && !sessionError;
     return (
@@ -150,6 +161,7 @@ export class SessionDetails extends React.Component<SessionDetailsProps, {}> {
                 <Button
                   disabled={session.active_queries?.length === 0}
                   onClick={() => {
+                    onTerminateStatementClick && onTerminateStatementClick();
                     if (session.active_queries?.length > 0) {
                       this.terminateQueryRef?.current?.showModalFor({
                         query_id: session.active_queries[0].id,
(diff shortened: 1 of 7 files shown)


koorosh (changes requested):
> Reviewed 5 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/ui/278)* status: 5 of 7 files reviewed, 2 unresolved discussions (waiting on @elkmaster and @nathanstilwell)

___
*[packages/cluster-ui/src/sessions/sessionDetails.tsx, line 60 at r1](https://reviewable.io/reviews/cockroachdb/ui/278#-MYczwDx3ZqQVrwl4-c_:-MYczwDx3ZqQVrwl4-ca:bz29th0) ([raw file](https://github.com/cockroachdb/ui/blob/00c4fcfc4741bea7ff326c7b3a862fa454503d02/packages/cluster-ui/src/sessions/sessionDetails.tsx#L60)):*
> ```XML
>   onBackButtonClick?: () => void;
>   onSessionActionClicked?: (
>     action: "Terminate Statement" | "Terminate Session",
> ```

@elkmaster , should component know about types of analytics actions? My concern is that this component won't allow to customize this callback prop to use with another analytics tool.
I suggest to define two separate callback functions in props - one is called when statement is terminated and another one when session is terminated.
___
*[packages/cluster-ui/src/sessions/sessionDetailsConnected.tsx, line 34 at r1](https://reviewable.io/reviews/cockroachdb/ui/278#-MYd0pmb7B7xV6cKrxbQ:-MYd0pmb7B7xV6cKrxbR:b-6o45ox) ([raw file](https://github.com/cockroachdb/ui/blob/00c4fcfc4741bea7ff326c7b3a862fa454503d02/packages/cluster-ui/src/sessions/sessionDetailsConnected.tsx#L34)):*
> ```XML
>       sessionError: state.adminUI.sessions.lastError,
>     }),
>     (dispatch: Dispatch) => ({
> ```

As far as you dispatch single action from a prop, you could use keep it as action creator object (without explicitly calling `dispatch`). It is necessary in case you need to dispatch two or more actions within single method.
___
*[packages/cluster-ui/src/sessions/sessionsPageConnected.tsx, line 45 at r1](https://reviewable.io/reviews/cockroachdb/ui/278#-MYd2J5G7VocBk0E2rBP:-MYd2J5G7VocBk0E2rBQ:b4bs4zf) ([raw file](https://github.com/cockroachdb/ui/blob/00c4fcfc4741bea7ff326c7b3a862fa454503d02/packages/cluster-ui/src/sessions/sessionsPageConnected.tsx#L45)):*
> ```XML
>       sessionsError: state.adminUI.sessions.lastError,
>     }),
>     (dispatch: Dispatch) => ({
> ```

The same as above, could be defined as action creator object
___
*[packages/cluster-ui/src/sessions/sessionsTable.tsx, line 74 at r1](https://reviewable.io/reviews/cockroachdb/ui/278#-MYd2nxkClB9uACO5Wx1:-MYd2nxkClB9uACO5Wx2:bn6m19c) ([raw file](https://github.com/cockroachdb/ui/blob/00c4fcfc4741bea7ff326c7b3a862fa454503d02/packages/cluster-ui/src/sessions/sessionsTable.tsx#L74)):*
> ```XML
>       <Link
>         onClick={() => {
>           onClick && onClick();
> ```

nit: `Link#onClick` method is optional, so it's possible to pass `onClick` function without validation:
```javascript
      <Link onClick={onClick} ... />
```

elkmaster (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/ui/278)* status: 2 of 7 files reviewed, 1 unresolved discussion (waiting on @koorosh and @nathanstilwell)

___
*[packages/cluster-ui/src/sessions/sessionDetails.tsx, line 60 at r1](https://reviewable.io/reviews/cockroachdb/ui/278#-MYczwDx3ZqQVrwl4-c_:-MYjR-9P3oZTINGPryaw:b-896fix) ([raw file](https://github.com/cockroachdb/ui/blob/00c4fcfc4741bea7ff326c7b3a862fa454503d02/packages/cluster-ui/src/sessions/sessionDetails.tsx#L60)):*
<details><summary><i>Previously, koorosh (Andrii Vorobiov) wrote…</i></summary><blockquote>

@elkmaster , should component know about types of analytics actions? My concern is that this component won't allow to customize this callback prop to use with another analytics tool.
I suggest to define two separate callback functions in props - one is called when statement is terminated and another one when session is terminated.
</blockquote></details>

Done.

koorosh (approved):
LGTM

PR #281, opened 2021-04-15
cluster-ui: clean up tooltip components [backport 20.2]

elkmaster:
before: in `cluster-ui` package we used legacy versions of
tootips based on antd and with a bunch of code duplicates and
hacks.

after: used `ui-components` tooltip component, removed unneeded
code and styles. also on process of work find out that barchats
legend tooltips missing styles - added them.

use tooltip component from `ui-components` library instead local
versions based on antd. to achieve this `ui-componets/Tooltip`
also needed some updates:
 - provide new style for table title usage (it have same color
   scheme as `default` but different typographic style)
 - add ability to hover on tooltip for clickable elements inside
 - other minor style updates and fixes

as side effect of this work there is a fix of dev process,
updates of storybook and webpack config was needed to run build
of `cluster-ui` with linked local `ui-components`.


nathanstilwell (approved):
LGTM

PR #282, opened 2021-04-19
cluster-ui: sessions style fixes [backport 20.2]

elkmaster:
updated css to fix corresponding style issues that was caused by
global css of managed service. disabled link for gateway node on
sessions details page.


koorosh (approved):
LGTM

PR #287, opened 2021-04-21
Update icon.md

Annebirzin:
# Docs // icons

Added icon guidelines and asset images


rsadres (approved):
looks great, I wish we could make the icons out of code, but that can be another phase

PR #290, opened 2021-04-22
ui-components: tooltip component updates

elkmaster:
provide new style for table title usage (it have same color
scheme as default but different typographic style)
add ability to hover on tooltip for clickable elements inside
other minor style updates and fixes
as side effect of this work there is a fix of dev process,
updates of storybook and webpack config was needed to run build
of cluster-ui with linked local ui-components.


packages/ui-components/src/Tooltip/Tooltip.tsx
```diff
@@ -2,6 +2,7 @@ import React, { FunctionComponent, ReactElement, useState } from "react";
 import { usePopper } from "react-popper";
 import classNames from "classnames/bind";
 import css from "./Tooltip.module.scss";
+import { debounce, throttle } from "lodash";
```
nathanstilwell, line 5: It seems like `debounce` and `throttle` aren't being used here.

packages/ui-components/src/Tooltip/Tooltip.tsx
```diff
@@ -67,36 +70,48 @@ export const Tooltip: FunctionComponent<TooltipProps> = ({
...
-        },
-        onMouseOut: () => {
-          if (visible) return;
-          popperElement.removeAttribute("data-show");
-        },
-      },
     );
   });
 
+  const timers: any = [];
```
nathanstilwell, line 76: We are trying to keep use of `any` out of the code base. You can replace it like this,

```suggestion
  const timers: Array<ReturnType<typeof setTimeout>> = [];
```

packages/ui-components/src/Tooltip/Tooltip.tsx
```diff
@@ -67,36 +70,48 @@ export const Tooltip: FunctionComponent<TooltipProps> = ({
...
+    <div
+      ref={setReferenceElement}
+      onMouseOver={() => {
+        const t = setTimeout(() => {
+          popperElement.setAttribute("data-show", "");
+        }, timeout);
+        timers.push(t);
+      }}
+      onMouseLeave={() => {
+        timers.forEach((t: any) => {
```
nathanstilwell, line 87: and you can replace this any with

```suggestion
       timers.forEach((t: ReturnType<typeof setTimeout>) => {
```

packages/ui-components/src/Tooltip/Tooltip.tsx
```diff
@@ -67,36 +70,48 @@ export const Tooltip: FunctionComponent<TooltipProps> = ({
...
+      }}
+    >
       {wrappedChildren}
 
       <div
         className={classnames}
         ref={setPopperElement}
         style={styles.popper}
         data-jest="tooltip"
+        onMouseLeave={e => {
```
nathanstilwell, line 99: `e` isn't used. I generally like to have it here to demonstrate that this is a callback that takes an _event_, but this will throw an linter warning so we might want to change it to
```suggestion
        onMouseLeave={() => {
```

nathanstilwell (changes requested):
Just a couple of quick changes that I hope aren't controversial.

nathanstilwell (approved):
LGTM

PR #293, opened 2021-04-22
cluster-ui: tooltip fixes [backport 20.2]

elkmaster:
before: in cluster-ui package we used legacy versions of
tootips based on antd and with a bunch of code duplicates and
hacks.

after: used ui-components tooltip component, removed unneeded
code and styles. also on process of work find out that barchats
legend tooltips missing styles - added them.


nathanstilwell (approved):
LGTM

PR #301, opened 2021-05-06
cluster-ui: return information about node regions to statement pages

maryliag:
Returning information from cache about which region each node is on.

Part of issue https://github.com/cockroachdb/cockroach/issues/59979
Dependency for PR (handles the backend changes) https://github.com/cockroachdb/cockroach/pull/64605

This will be followed by another PR with the changes on the ui visible to the user.

Release note: None

### DIFF
--- a/packages/cluster-ui/src/store/nodes/nodes.selectors.ts
+++ b/packages/cluster-ui/src/store/nodes/nodes.selectors.ts
@@ -4,6 +4,8 @@ import { AppState } from "../reducers";
 import { getDisplayName } from "../../nodes";
 import { livenessStatusByNodeIDSelector } from "../liveness";
 import { accumulateMetrics } from "../../util";
+import { cockroach } from "@cockroachlabs/crdb-protobuf-client";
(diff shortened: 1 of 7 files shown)


packages/cluster-ui/src/store/nodes/nodes.selectors.ts
```diff
@@ -29,3 +31,24 @@ export const nodeDisplayNameByIDSelector = createSelector(
     return result;
   },
 );
+
+export function getRegionFromLocality(locality: ILocality): string {
+  for (let i = 0; i < locality.tiers.length; i++) {
+    if (locality.tiers[i].key === "region") return locality.tiers[i].value;
+  }
+  return "";
```
Azhng, line 39: When would there be a case where this information is not available?

maryliag: when the locality is not setup, for example. Even during the setup some queries might run without the value yet.
Also, we always want to function to have a backup value to return in case something is wrong.

Azhng (approved):
LGTM

PR #302, opened 2021-05-06
cluster-ui: adding information about node and region on statement details

maryliag:
Add information about which node and region a statement was executed on.

Screen Shot 2021-05-06 at 6 10 48 PM

Resolves: https://github.com/cockroachdb/cockroach/issues/59979

Release note (ui change): Display on Statement Details information about
nodes and regions a statement was executed on.

### DIFF
--- a/packages/cluster-ui/src/statementDetails/statementDetails.tsx
+++ b/packages/cluster-ui/src/statementDetails/statementDetails.tsx
@@ -14,6 +14,7 @@ import {
   Bytes,
   Duration,
   FixLong,
+  longToInt,
   appAttr,
   NumericStat,
   StatementStatistics,
@@ -21,6 +22,7 @@ import {
   getMatchParamByName,
   formatNumberForDisplay,
   calculateTotalWorkload,
+  unique,
 } from "src/util";
 import { Loading } from "src/loading";
 import { Button } from "src/button";
@@ -33,7 +35,6 @@ import {
   approximify,
   latencyBreakdown,
   genericBarChart,
-  longToInt,
   formatTwoPlaces,
 } from "src/barCharts";
 import {
@@ -156,6 +157,17 @@ function AppLink(props: { app: string }) {
       to={`/statements/${encodeURIComponent(props.app)}`}
     >
       {props.app}
+    </Link>
+  );
+}
+
+function NodeLink(props: { node: string }) {
+  return (
+    <Link
+      className={cx("app-name")}
+      to={`/node/${encodeURIComponent(props.node)}`}
+    >
+      N{props.node}
     </Link>
   );
 }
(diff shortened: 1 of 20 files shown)


packages/cluster-ui/src/util/appStats/appStats.ts
```diff
@@ -107,6 +107,17 @@ export function addExecStats(a: ExecStats, b: ExecStats): Required<ExecStats> {
   };
 }
 
+function combinesUnique<T>(a: Array<T>, b: Array<T>): Array<T> {
```
nathanstilwell, line 110: I like to use `Set` to create unique arrays. You may consider doing something like,

```js
const combineUnique = <T>(a: Array<T>, b: Array<T>): Array<T> => [...new Set([...a, ...b])];
```

This could also be added to `src/utils`.

.... On second thought, I think I would rather have a `unique` utility function,

```js
const uniq = <T, >(a: T[]): T[] => [...new Set([...a])];
```

And then use this function to return a unique concatenation of  multiple arrays. Like so,

`nodes: unique([...a.nodes, ...b.nodes]),`

nathanstilwell (approved):
I have one minor suggestion for _combining arrays into a unique array_, but otherwise this looks good.

packages/cluster-ui/src/statementDetails/statementDetails.tsx
```diff
@@ -435,6 +447,15 @@ export class StatementDetails extends React.Component<
 
     const statsByNode = this.props.statement.byNode;
     const totalWorkload = calculateTotalWorkload(statsByNode);
+    const nodes = stats.nodes
+      ? stats.nodes.sort().map(node => node.toString())
+      : [];
+    const regions: string[] = [];
+    stats.nodes.forEach(node => {
+      if (!regions.includes(nodeRegions[node.toString()]))
+        regions.push(nodeRegions[node.toString()]);
```
Azhng, line 456: small nit: `nodeRegions[node.toString()]` is being used twice here. Perhaps store it in a variable to prevent two function invocation and dict lookup. Though `set` might be a more suitable data structure here.

packages/cluster-ui/src/statementDetails/statementDetails.tsx
```diff
@@ -541,6 +562,33 @@ export class StatementDetails extends React.Component<
...
+                      nodes.map(n => <NodeLink node={n} key={n} />),
+                      ", ",
+                    )}
+                  </Text>
+                </div>
+                <div className={summaryCardStylesCx("summary--card__item")}>
+                  <Text>Regions</Text>
+                  <Text>
+                    {intersperse<ReactNode>(
+                      regions.map(r => r),
```
Azhng, line 578: Is this `map` necessary?

packages/cluster-ui/src/statementDetails/statementDetails.tsx
maryliag: done

PR #311, opened 2021-05-13
cluster-ui: remove tagname css rules

elkmaster:
antd style set contains a bunch of direct rules for tagnames.
this leads to style conflicts when we use cluster-ui in
managed service env. to avoid this only fast solution was
to copy antd styles in src and comment out all rules of
such type.

### DIFF
--- a/packages/cluster-ui/babel.config.js
+++ b/packages/cluster-ui/babel.config.js
@@ -19,7 +19,6 @@ const plugins = [
   "@babel/proposal-object-rest-spread",
   // @babel/plugin-transform-runtime is required to support dynamic loading of cluster-ui package
   "@babel/plugin-transform-runtime",
-  ["import", { "libraryName": "antd", "style": "css" }],
 ];
 
 const env = {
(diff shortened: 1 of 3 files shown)


nathanstilwell:
Don't the global styles in CockroachCloud console still conflict with the Cluster UI components since those AntD styles are in the global scope? It's not immediately obvious how this change resolves the issue.

elkmaster:
there is such order of css apply in managed service:
 - import antd.css
 - import other managed service css rules, with antd rewrites (this is unavoidable due to antd lib design)
 - lazy load cluster ui, that contain antd css again
 
so basically we add one more antd css set of rules after it was rewritten in managed service. it is not a problem for class\id based selectors but reset direct tag name selectors. 

current change fix at least problem with headers mentioned in CRDB-6956,
to fix all potential problems with global styles we need to eliminate usage of antd.

nathanstilwell (approved):
LGTM

nathanstilwell:
@elkmaster This probably needs to be backported to `cluster-ui-21.1` now.

elkmaster:
backports:
https://github.com/cockroachdb/ui/pull/323
https://github.com/cockroachdb/ui/pull/324

PR #312, opened 2021-05-13
cluster-ui: adding information about node and region on stmt/transaction pages

maryliag:
Display information about which Regions/Nodes a statement was executed and
a filter option for region and another for nodes.
Both new filters will only be displayed if there is more than one value.

**Values on the table:**
Screen Shot 2021-05-13 at 6 30 56 PM

**Filter:**
Screen Shot 2021-05-13 at 6 31 40 PM

**Columns selector on statement page:**
Screen Shot 2021-05-13 at 6 31 16 PM

Release note (ui change): Display on Statement and Transaction pages information
about nodes and regions a statement was executed on.

### DIFF
--- a/packages/cluster-ui/src/barCharts/utils.ts
+++ b/packages/cluster-ui/src/barCharts/utils.ts
@@ -1,5 +1,6 @@
 import { format as d3Format } from "d3-format";
 import * as protos from "@cockroachlabs/crdb-protobuf-client";
+import { TransactionInfo } from "../transactionsTable";
 
 type StatementStatistics = protos.cockroach.server.serverpb.StatementsResponse.ICollectedStatementStatistics;
 type Transaction = protos.cockroach.server.serverpb.StatementsResponse.IExtendedCollectedTransactionStatistics;
@@ -10,7 +11,7 @@ export const formatTwoPlaces = d3Format(".2f");
 
 export function bar(
   name: string,
-  value: (d: StatementStatistics | Transaction) => number,
+  value: (d: StatementStatistics | Transaction | TransactionInfo) => number,
 ) {
   return { name, value };
 }
--- a/packages/cluster-ui/src/queryFilter/filter.module.scss
+++ b/packages/cluster-ui/src/queryFilter/filter.module.scss
@@ -56,7 +56,7 @@ $dropdown-hover-color: darken($colors--background, 2.5%);
       }
 
       &__margin-top {
-        margin-top: 27px;
+        margin-top: 15px;
       }
     }
 
--- a/packages/cluster-ui/src/queryFilter/filter.tsx
+++ b/packages/cluster-ui/src/queryFilter/filter.tsx
@@ -23,9 +23,13 @@ interface QueryFilter {
   activeFilters: number;
   filters: Filters;
   dbNames?: string[];
+  regions?: string[];
+  nodes?: string[];
   showDB?: boolean;
   showSqlType?: boolean;
   showScan?: boolean;
+  showRegions?: boolean;
+  showNodes?: boolean;
 }
 interface FilterState {
   hide: boolean;
@@ -45,6 +49,8 @@ export interface Filters {
   sqlType?: string;
   fullScan?: boolean;
   distributed?: boolean;
+  regions?: string;
+  nodes?: string;
 }
 
 const timeUnit = [
@@ -59,6 +65,8 @@ export const defaultFilters: Filters = {
   fullScan: false,
   sqlType: "",
   database: "",
+  regions: "",
+  nodes: "",
 };
 
 /**
@@ -101,6 +109,8 @@ export const inactiveFiltersState: Filters = {
   fullScan: false,
   sqlType: "",
   database: "",
+  regions: "",
+  nodes: "",
 };
 
 export const calculateActiveFilters = (filters: Filters) => {
@@ -213,10 +223,14 @@ export class Filter extends React.Component<QueryFilter, FilterState> {
     const {
       appNames,
       dbNames,
+      regions,
+      nodes,
       activeFilters,
       showDB,
       showSqlType,
       showScan,
+      showRegions,
+      showNodes,
     } = this.props;
     const dropdownArea = hide ? hidden : dropdown;
     const customStyles = {
@@ -254,7 +268,6 @@ export class Filter extends React.Component<QueryFilter, FilterState> {
           isSelected: this.isOptionSelected(db, filters.database),
         }))
       : [];
-
     const databaseValue = databasesOptions.filter(option => {
       return filters.database.split(",").includes(option.label);
     });
@@ -267,6 +280,52 @@ export class Filter extends React.Component<QueryFilter, FilterState> {
           field="database"
           parent={this}
           value={databaseValue}
+        />
+      </div>
+    );
+
+    const regionsOptions = showRegions
+      ? regions.map(region => ({
+          label: region,
+          value: region,
+          isSelected: this.isOptionSelected(region, filters.regions),
+        }))
+      : [];
+    const regionsValue = regionsOptions.filter(option =>
+      filters.regions.split(",").includes(option.label),
+    );
+    const regionsFilter = (
+      <div>
+        <div className={filterLabel.margin}>Region</div>
+        <MultiSelectCheckbox
+          options={regionsOptions}
+          placeholder="All"
+          field="regions"
+          parent={this}
+          value={regionsValue}
+        />
+      </div>
+    );
+
+    const nodesOptions = showNodes
+      ? nodes.map(node => ({
+          label: node,
+          value: node,
+          isSelected: this.isOptionSelected(node, filters.nodes),
+        }))
+      : [];
+    const nodesValue = nodesOptions.filter(option => {
+      return filters.nodes.split(",").includes(option.label);
+    });
+    const nodesFilter = (
+      <div>
+        <div className={filterLabel.margin}>Node</div>
+        <MultiSelectCheckbox
+          options={nodesOptions}
+          placeholder="All"
+          field="nodes"
+          parent={this}
+          value={nodesValue}
         />
       </div>
     );
@@ -297,7 +356,6 @@ export class Filter extends React.Component<QueryFilter, FilterState> {
     const sqlTypeValue = sqlTypes.filter(option => {
       return filters.sqlType.split(",").includes(option.label);
     });
-
     const sqlTypeFilter = (
       <div>
         <div className={filterLabel.margin}>Statement Type</div>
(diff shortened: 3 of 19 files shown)


packages/cluster-ui/src/transactionsPage/utils.ts
```diff
@@ -69,38 +81,94 @@ export const filterTransactions = (
...
+export const generateRegionNode = (
+  transaction: Transaction,
+  statements: Statement[],
+  nodeRegions: { [p: string]: string },
+): string[] => {
+  const regions: { [region: string]: Set<number> } = {};
+  getStatementsById(transaction.stats_data.statement_ids, statements).forEach(
+    stmt => {
+      stmt.stats.nodes.forEach(node => {
+        if (Object.keys(regions).includes(nodeRegions[node.toString()])) {
```
Azhng, line 148: nit: would be nice to extract `nodeRegions[node.toString()]` to its own variable since it's used multiple times in this func.

packages/cluster-ui/src/barCharts/utils.ts
```diff
@@ -1,5 +1,6 @@
 import { format as d3Format } from "d3-format";
 import * as protos from "@cockroachlabs/crdb-protobuf-client";
+import { TransactionInfo } from "../transactionsTable";
```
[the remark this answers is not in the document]
maryliag: Good question, I don't think we do

packages/cluster-ui/src/sortedtable/sortedtable.tsx
```diff
@@ -370,3 +371,30 @@ export class SortedTable<T> extends React.Component<
...
+ * E.g. `value1, value2, value3` with maxLength 10 will display
+ * `value1, va...` on the table and
+ * `value1
+ * value2
+ * value3`  on the tooltip.
+ * @param value a string with elements separated by `, `.
+ * @param maxLength the max length to which it should display value
+ * and hide the remaining.
+ */
+export function longListWithTooltip(value: string, maxLength: number) {
```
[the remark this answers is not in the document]
maryliag: this is a function specific to be used on sortedTables when adding longList, if in the future it's used for other places we might consider a change in location, but for now I want the scope to be only this place

packages/cluster-ui/src/sortedtable/sortedtable.tsx
```diff
@@ -370,3 +371,30 @@ export class SortedTable<T> extends React.Component<
...
+  const summary =
+    value.length > maxLength ? value.slice(0, maxLength) + "..." : value;
+  return (
+    <Tooltip
+      placement="bottom"
+      content={
+        <pre className={cx("break-line")}>{value.split(", ").join("\r\n")}</pre>
```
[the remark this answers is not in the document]
maryliag: That's because some browsers use `\n`, some `\r`, some `\r\n`. So using `\r\n` you make sure it works on all :)

packages/cluster-ui/src/transactionsPage/transactions.fixture.ts
```diff
@@ -460,6 +467,94 @@ export const data: cockroach.server.serverpb.IStatementsResponse = {
       },
       id: Long.fromInt(106),
     },
+    {
```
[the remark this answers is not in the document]
maryliag: The id for this statement is 107, but for some reason it was added out or order with the other statements, I just changed so it's in order compared to the others, there is no change on the values itself

packages/cluster-ui/src/queryFilter/filter.tsx
```diff
@@ -271,6 +285,53 @@ export class Filter extends React.Component<QueryFilter, FilterState> {
...
+          label: region,
+          value: region,
+          isSelected: this.isOptionSelected(region, filters.regions),
+        }))
+      : [];
+
+    const regionsValue = regionsOptions.filter(option =>
+      filters.regions.split(",").includes(option.label),
+    );
+    const regionsFilter = (
```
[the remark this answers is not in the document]
maryliag: I like to keep constant creations (you can notice these are all constants creations, not functions) that are very related in groups, so creating all constants for regions + space + constants for nodes + space and so on.

packages/cluster-ui/src/statementsTable/statementsTable.tsx
```diff
@@ -250,5 +268,39 @@ export function makeNodesColumns(
...
-  return original.concat(makeCommonColumns(statements, totalWorkload));
+  return original.concat(
+    makeCommonColumns(statements, totalWorkload, nodeRegions),
+  );
+}
+
+// For each statement, generate the list of regions and nodes it was
+// executed on.
+// E.g. of one element of the list: `gcp-us-east1 (n1, n2, n3)`
+export function generateRegionNodeForStatements(
```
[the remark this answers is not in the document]
maryliag: done

packages/cluster-ui/src/transactionsPage/utils.ts
maryliag: done

packages/cluster-ui/src/queryFilter/filter.tsx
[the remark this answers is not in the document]
Azhng: I see. Upon seeing the `=>` my monkey brain just instantly thought these are functions 😅

Azhng (commented):
LGTM. It'd be great if someone from cluster-ui-prs can take a look at as well.

packages/cluster-ui/src/queryFilter/filter.module.scss
```diff
@@ -56,7 +56,7 @@ $dropdown-hover-color: darken($colors--background, 2.5%);
       }
 
       &__margin-top {
-        margin-top: 27px;
+        margin-top: 15px;
```
j-low, line 59: Not flagging necessarily, but do we use any tokenization for gutter values?

packages/cluster-ui/src/sortedtable/sortedtable.module.scss
```diff
@@ -97,6 +97,10 @@
 
 }
 
+.break-line {
+    white-space:pre-wrap !important;
```
j-low, line 101: It would be ideal if we can find a way around using an `!important` flag.

packages/cluster-ui/src/statementsPage/statementsPage.tsx
```diff
@@ -306,6 +320,25 @@ export class StatementsPage extends React.Component<
...
+            regions,
+          ),
+      )
+      .filter(
+        statement =>
+          nodes.length == 0 ||
+          containAny(
+            statement.stats.nodes.map(node => "n" + node),
+            nodes,
+          ),
```
j-low, line 341: I like these pithy functional statements quite a lot but it would also be nice to perhaps have an explanatory comment somewhere summarizing what filtering is taking place and for what purpose.  This will help make it easier to extend or modify in the future.

packages/cluster-ui/src/statementsPage/statementsPage.tsx
```diff
@@ -330,18 +364,32 @@ export class StatementsPage extends React.Component<
...
       onDiagnosticsReportDownload,
       onStatementClick,
     );
 
+    // If it's multi-region, we want to show the Regions/Nodes column by default
+    // and hide otherwise.
+    if (regions.length > 1) {
+      columns.filter(c => c.name === "regionNodes")[0].showByDefault = true;
+    }
+
```
j-low, line 392: Re: my comment above about adding comments, I think this is a nice example to replicate in similar instances.

packages/cluster-ui/src/statementsTable/statementsTable.tsx
```diff
@@ -250,5 +268,39 @@ export function makeNodesColumns(
...
+) {
+  statements.forEach(stmt => {
+    const regions: { [region: string]: Set<number> } = {};
+    stmt.stats.nodes.forEach(node => {
+      if (Object.keys(regions).includes(nodeRegions[node.toString()])) {
+        regions[nodeRegions[node.toString()]].add(longToInt(node));
+      } else {
+        regions[nodeRegions[node.toString()]] = new Set([longToInt(node)]);
+      }
+    });
```
j-low, line 291: A short comment summary of this block would be helpful as well.

packages/cluster-ui/src/transactionsPage/utils.ts
```diff
@@ -69,38 +81,95 @@ export const filterTransactions = (
...
+  const regions: { [region: string]: Set<number> } = {};
+  getStatementsById(transaction.stats_data.statement_ids, statements).forEach(
+    stmt => {
+      stmt.stats.nodes.forEach(n => {
+        const node = n.toString();
+        if (Object.keys(regions).includes(nodeRegions[node])) {
+          regions[nodeRegions[node]].add(longToInt(n));
+        } else {
+          regions[nodeRegions[node]] = new Set([longToInt(n)]);
+        }
```
j-low, line 153: Ditto for comment on this section.

j-low (changes requested):
Great stuff, thanks for all the work!  I have a few requests re: commenting as well as once about the use of the `!important` flag in the CSS but otherwise this looks good.

packages/cluster-ui/src/queryFilter/filter.module.scss
maryliag: We use tokens for sizes of text, button, things like that. When it's a very specific space for a specific component I use the value itself, because it's very unlikely this value will be aligned with any other components

packages/cluster-ui/src/sortedtable/sortedtable.module.scss
maryliag: I did try other option without using it, but none of them were working unfortunately. If you have any ideas, I'm open to it!

packages/cluster-ui/src/statementsPage/statementsPage.tsx
maryliag: done

packages/cluster-ui/src/statementsPage/statementsPage.tsx
maryliag: Got it! :)

packages/cluster-ui/src/statementsTable/statementsTable.tsx
maryliag: done

packages/cluster-ui/src/transactionsPage/utils.ts
maryliag: done

j-low (approved):

j-low:
@maryliag changes approved per our offline convo 👍

PR #319, opened 2021-05-20
21.2 Version alignment

nathanstilwell:
# CLUSTER-UI // Minor version increase for CRDB 21.2 development

Increasing the minor version of cluster-ui to represent the next
unreleased version of CRDB (21.2) that is being developed on the master
branch of `cockroachdb/cockroach`.

There was no mention of how minor versions of cluster-ui are aligned to
major release versions of CRDB, so adding a section outlining this
mapping. Also documenting the versioning skew and deprecation that
happened after the CRDB 21.1 release.


koorosh (approved):
LGTM

PR #321, opened 2021-05-21
cluster-ui: fix styling for statements filter component

koorosh:
Filter component is based on `react-select` package and has been moved
form cockroachdb repository initially. But except component related
styles, this component relied on global styles defined in `cockroachdb/cockroach/pkg/ui/styl/shame.styl`
 file and wasn't moved with component.
 Current fix imports default `react-select` component styles and
 overridden styles as well.

Screen Shot 2021-05-21 at 5 59 03 PM

Screen Shot 2021-05-21 at 5 43 47 PM


koorosh:
This fix should be applied for in cluster-ui-21.1 release branch only.

PR #323, opened 2021-05-26
backport[20.2] cluster-ui: remove tagname css rules

elkmaster:
antd style set contains a bunch of direct rules for tagnames.
this leads to style conflicts when we use cluster-ui in
managed service env. to avoid this only fast solution was
to copy antd styles in src and comment out all rules of
such type.


nathanstilwell (approved):
LGTM

PR #344, opened 2021-07-29
ui-components: tooltip imporvements

elkmaster:
before:
- tooltip hover area and tooltip itself was
clickable, and in some cases it could propagate
event to parent and lead to unwanted actions.
- tooltip could stuck visible if page scrolled

after:
- disable pointer event on hover area, stop
propagation on tooltip area.
- hide tooltip on window scroll

additional changes: disable cluster-ui build via
renaming it package.json file. fix icons pkg build,
was corrupted due to renovate bot pin modules
versions


packages/ui-components/src/Tooltip/Tooltip.tsx
```diff
@@ -64,11 +64,7 @@ export const Tooltip: FunctionComponent<TooltipProps> = ({
 
   const wrappedChildren = React.Children.map(children, child => {
     return React.cloneElement(
-      typeof children === "string" ? (
-        <span>{child}</span>
-      ) : (
-        (child as ReactElement)
-      ),
+      typeof children === "string" ? <>{child}</> : (child as ReactElement),
```
Santamaura, line 67: what is the reason for switching from span to empty tag; Is it just because it's more compact and no attr are added to it?

Santamaura (commented):
Rest 👍

packages/ui-components/src/Tooltip/Tooltip.tsx
elkmaster: just to avoid creation unneeded dom element

koorosh:
tried out it with storybook and noticed that tooltips don't hide when you simply move out cursor

Screen Shot 2021-08-03 at 11 58 14 AM

elkmaster:
@koorosh, thanks, added fix for that:
https://github.com/cockroachdb/ui/pull/344/files#diff-f78af0a31de2edadb9ecd5ac3f5ed5999c1f9a66965d75b74da494b7d3c8ea49R97

koorosh (approved):
Nice! LGTM

PR #372, opened 2021-08-10
remove unnecessary nested yarn.lock files

koorosh:
In Yarn monorepo project, only one yarn.lock file should
exist on the top level of the project.
For some reason, unnecessary lock files in workspaces
weren't removed.
This could cause inconsistency when yarn commands
ran from workspaces level or could confuse renovate bot
to properly detect monorepo structure and update lock
file along with `package.json` file


nathanstilwell (approved):
I'm into it.

PR #416, opened 2021-12-16
design-tokens: export stylus tokens

jocrl:
Export stylus tokens for use in db console
Motivated by https://github.com/cockroachdb/cockroach/issues/72674


laurenbarker (approved):
Thank you!

PR #419, opened 2021-12-30
CODEOWNERS // Formatting to match CRL standard

nathanstilwell:
Bringing CODEOWNERS into alignment with the owners file in CRDB and
Managed Service. Team has been created in the @cockroachlabs
organization.


laurenbarker (approved):
Thanks for updating this!

nathanstilwell:
One minor update. I'm not sure if you can set the owner to a team outside of the organization. This is something we had to do for Cluster UI as well. So CRUX will have a team in both @cockroachlabs and @cockroachdb.

PR #425, opened 2022-01-07
ui-components: remove astroturf from dependencies

laurenbarker:
We were experimenting with CSS-in-JS libraries when ui-components was
first created. astroturf would allow us to introduce CSS-in-JS, but
continue to use our existing CSS processing pipeline (i.e., SCSS).

However, we aren't using CSS-in-JS in any of our files, so we can
remove the dependency.

#### Checklist
- [x] I have written or updated test for the changes I made (N/A)
- [x] I have updated the README of the package I'm working in to reflect my changes (N/A; no docs mention astroturf)
- [x] I have added or updated Storybook if appropriate for my changes (N/A)

### DIFF
--- a/packages/ui-components/package.json
+++ b/packages/ui-components/package.json
@@ -52,7 +52,6 @@
     "@types/react": "16.14.21",
     "@typescript-eslint/eslint-plugin": "2.34.0",
     "@typescript-eslint/parser": "2.34.0",
-    "astroturf": "0.10.5",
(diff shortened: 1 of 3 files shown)


nathanstilwell (approved):
LGTM

sjbarag (approved):
🎉 great catch!

laurenbarker:
TFTRs!

PR #430, opened 2022-01-11
WIP: Attempting to test Color in React Testing Library

nathanstilwell:
I thought that I could simply compare the color of the rendered
component to the expected color of our design tokens. The issue
that I'm currently stuck on is that Jest won't transpile any
dependencies (meaning that if I import the existing tokens Jest will
fail, not recognizing `export`). If I compile the tokens to a CommonJS
module then TypeScript will fail the test saying there are no type
definitions.

Beyond that, is this even a strategy that we like? How else could be
replace a test for appropriate intent?


laurenbarker:
>Beyond that, is this even a strategy that we like? How else could be
replace a test for appropriate intent?

Visual snapshot tests would give us the most confidence that our styles are being applied correctly. In this repo, we could use storybook to render the components in different states and then take snapshots of the pages for testing. Percy makes this really easy https://docs.percy.io/docs/storybook

nathanstilwell:
@laurenbarker I like your suggestion of visual snapshots paired with React Testing Library. I simply added a test to check that the correct class is being applied based on `intent`. In the process I cleaned up the types a little bit (`InlineAlert` was using `IconIntent` which introduced some weirdness and some type errors).

packages/ui-components/jest.testing.config.js
```diff
@@ -5,9 +5,7 @@ module.exports = {
     "\\.(css|scss)$": "identity-obj-proxy",
   },
   roots: ["<rootDir>/src"],
-  setupFilesAfterEnv: [
-    "@testing-library/jest-dom/extend-expect",
-  ],
+  setupFilesAfterEnv: ["@testing-library/jest-dom/extend-expect"],
```
laurenbarker, line 8: nit: This file could be excluded from this PR since this diff is from previous changes that were reverted.

yarn.lock
```diff
@@ -2819,7 +2819,7 @@
     minimist "^1.2.0"
 
 "@cockroachlabs/ui-components@link:packages/ui-components":
-  version "0.2.24-alpha.3"
+  version "0.2.24-alpha.4"
```
laurenbarker, line 2822: nit: This file could be excluded from this PR since this diff is from previous changes that were reverted.

packages/ui-components/src/InlineAlert/InlineAlert.tsx
```diff
@@ -2,13 +2,15 @@ import React, { useMemo } from "react";
...
+import { Icon } from "../Icon/Icon";
 import { Text } from "../Typography";
 
 const cx = classNames.bind(styles);
 
+export type InlineAlertIntent = "danger" | "info" | "success" | "warning";
+
 interface InlineAlertBaseProps {
-  intent?: IconIntent;
+  intent?: InlineAlertIntent;
```
laurenbarker, line 13: Good catch!

laurenbarker (commented):
Thanks for updating the types!

PR #442, opened 2022-03-18
eslint-config, ui-components: add stricter rules to eslint config

laurenbarker:
We have been adding stricter rules to the eslint config in the
managed-service repo. Now that the eslint version has been updated, we
can update the rules as well.

There was one naming rule violation in the ui-components package that
is now ignored because the we are passing a component in as a prop. The
casing is correct.

I moved everything below https://github.com/cockroachlabs/managed-service/blob/master/console/.eslintrc.json#L6.


sjbarag (approved):
Excellent! 🎉

laurenbarker:
TFTR! So fast!

PR #454, opened 2022-05-24
Add ExpandableText component to ui-components

jocrl:
This commit adds an ExpandableText component to ui-components.

# ui-components // Add ExpandableText component

Collapsed:
image

Expanded:
image

Does not exceed length:
image

This component was added for https://github.com/cockroachdb/cockroach/pull/81180#issuecomment-1134899951, for the Job details error table.
It is being added in this repo per [this conclusion](https://cockroachlabs.slack.com/archives/C3T5LG9R8/p1645029579140249?thread_ts=1644965483.606339&cid=C3T5LG9R8):

> So, tl;dr for using db-console/cluster-ui components in managed-service:
> - For components (e.g. SQL Activity pages) that are specific to a database version and its APIs, they should live in cluster-ui and be imported in managed-service using componentFactory. This does a dynamic import that allows cluster-ui to be bundled separately from the main managed-service bundle
> - For generic components (e.g. Tooltips, SqlBox) that are not tied to a specific database version, they should ideally be moved to https://github.com/cockroachdb/ui. This allows them to be shared across db-console, cluster-ui, and managed-service
>   - As a fallback alternative, the code for these components can be duplicated in the managed-service repo, at the cost of code duplication

#### Checklist

### DIFF
--- a/packages/ui-components/src/ExpandableText/ExpandableText.module.scss
+++ b/packages/ui-components/src/ExpandableText/ExpandableText.module.scss
@@ -1,0 +1,10 @@
+@import "../styles/tokens.scss";
+
+.anchor {
+  color: $color-font-link-base;
+  fill: $color-font-link-base;
+  &:hover {
+    color: $color-font-link-hover;
+    fill: $color-font-link-hover;
+  }
+}
--- a/packages/ui-components/src/ExpandableText/ExpandableText.tsx
+++ b/packages/ui-components/src/ExpandableText/ExpandableText.tsx
@@ -1,0 +1,65 @@
+import React, { useEffect, useState } from "react";
+import styles from "./ExpandableText.module.scss";
+import classNames from "classnames/bind";
+import LinesEllipsis from "react-lines-ellipsis";
+
+interface ExpandableTextProps {
+  text: string;
+  maxLine: number;
+}
+
+const cx = classNames.bind(styles);
+
+export function ExpandableText({
+  text,
+  maxLine,
+}: ExpandableTextProps): React.ReactElement {
+  const [isExpanded, setIsExpanded] = useState(false);
+
+  useEffect(() => {
+    // When the text changes, reset to the default of isExpanded=false.
+    // This is needed in case the text changes without re-mounting the component.
+    setIsExpanded(false);
+  }, [text]);
+
+  return (
+    <>
+      {!isExpanded && (
+        <LinesEllipsis
+          text={text}
+          maxLine={maxLine}
+          /**
+           * The type definition for the ellipsis prop (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-lines-ellipsis/index.d.ts#L14)
+           * does not match the component behavior or the docs (https://github.com/xiaody/react-lines-ellipsis#propsellipsis-node),
+           * which says that ellipsis takes type Node.
+           */
+          ellipsis={
+            (
+              <>
+                {"… "}
+                <a
+                  className={cx("crl-anchor")}
+                  onClick={() => {
+                    setIsExpanded(true);
+                  }}
+                >
+                  Show More
+                </a>
+              </>
+            ) as unknown as string
+          }
+        />
+      )}
+      {isExpanded && (
+        <div>
+          {text}{" "}
+          <a className={cx("anchor")} onClick={() => setIsExpanded(false)}>
+            Show Less
+          </a>
+        </div>
+      )}
+    </>
+  );
+}
+
+export default ExpandableText;
--- a/packages/storybook-ui-components/stories/ExpandableText.stories.tsx
+++ b/packages/storybook-ui-components/stories/ExpandableText.stories.tsx
@@ -1,0 +1,25 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+
+import { ExpandableText } from "@cockroachlabs/ui-components";
+
+const divWidth = "200px";
+const fourLinetext =
+  "Very very very very very very very very very very very very very very very very very very very long text.";
+
+storiesOf("ExpandableText", module)
+  .add("Default", () => (
+    <div style={{ width: divWidth }}>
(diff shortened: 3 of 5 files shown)


Annebirzin:
Nice! Can we just confirm it's using out core blue `#0055FF` ?

jocrl:
@Annebirzin updated, thanks for the catch!

maryliag (approved):
LGTM!

packages/ui-components/src/ExpandableText/ExpandableText.tsx
```diff
@@ -0,0 +1,46 @@
...
+    setIsExpanded(false);
+  }, [text]);
+
+  if (text.length <= characterLimit) return <>{text}</>;
+
+  return (
+    <>
+      {!isExpanded && (
+        <>
+          {text.slice(0, characterLimit) + "..."}{" "}
```
sjbarag, line 28: nit: any chance we could use an ellipsis character (…) instead of three periods?

sjbarag (changes requested):
The example shown in https://github.com/cockroachdb/cockroach/pull/81180#issuecomment-1134899951 is a multi-line element, but the `ExpandableText` element here is based on pure character count.  Since proportional fonts will produce different widths of text with the same number of characters (think "iii" vs "WWW"), I'm not sure how we'll be able to accurately limit a component to two rendered lines this way 🤔 

Is this something we could accomplish by adding/removing `text-overflow: ellipsis` (https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow) to the text we want expanded, rather than using character counts?

sjbarag:
> The example shown in [cockroachdb/cockroach#81180 (comment)](https://github.com/cockroachdb/cockroach/pull/81180#issuecomment-1134899951) is a multi-line element, but the `ExpandableText` element here is based on pure character count. Since proportional fonts will produce different widths of text with the same number of characters (think "iii" vs "WWW"), I'm not sure how we'll be able to accurately limit a component to two rendered lines this way 🤔
> 
> Is this something we could accomplish by adding/removing `text-overflow: ellipsis` (https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow) to the text we want expanded, rather than using character counts?

Right, so CSS `text-overflow: ellipsis` only supports single-line overflows.  Using multiple lines would require either `-webkit-line-clamp` (which won't work on IE11: https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp) or a JS solution like https://www.npmjs.com/package/react-lines-ellipsis or the version provided by `antd` (https://ant.design/components/typography/#components-typography-demo-ellipsis; though IIRC we're trying to avoid antd in this repo?)

Annebirzin (approved):

jocrl:
Ah, thanks Sean! I found antd's Expand, but hadn't found the ellipsis.

I initially put this in this repo to hopefully contribute to a cleaner unified UI state, but if there's concerns about having antd in this repo I'll ask that question on #frontend-learners and perhaps just put this component in cluster-ui for now + an issue for the future. I'll want to wrap the antd component in our own in order to have the Show More/Show Less, so it's not just an issue of directly using an antd component.

packages/ui-components/src/ExpandableText/ExpandableText.module.scss
```diff
@@ -0,0 +1,10 @@
+@import "../styles/tokens.scss";
+
+.crl-anchor {
```
nathanstilwell, line 3: Since this is a CSS module, you don't need the `crl-` prefix for this classname.

jocrl:
Modified the implementation to truncate by lines instead of character count. Removed the test, since RTL doesn't seem to correctly take the fixed width into account when rendering.
Also removed the `crl-` CSS prefix.

Thanks for the feedback; PTAL!

nathanstilwell (commented):
A few comments ...

packages/ui-components/src/ExpandableText/ExpandableText.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+const cx = classNames.bind(styles);
+
+export function ExpandableText({
+  text,
+  maxLine,
+}: ExpandableTextProps): React.ReactElement {
+  const [isExpanded, setIsExpanded] = useState(false);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
sjbarag: > but wouldn't that happen anyway on a re-render (since the default state is false)?

IIRC the default value is only applied when the component is mounted, so an effect is necessary here I think!

packages/ui-components/src/ExpandableText/ExpandableText.tsx
```diff
@@ -0,0 +1,60 @@
...
+    setIsExpanded(false);
+  }, [text]);
+
+  return (
+    <>
+      {!isExpanded && (
+        <LinesEllipsis
+          text={text}
+          maxLine={maxLine}
+          ellipsis={
```
jocrl, line 31: Thanks everyone!

@nathanstilwell, yeah the [type definition](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-lines-ellipsis/index.d.ts#L14) for the `ellipsis` prop is not up-to-date with the docs, and this is generating a typescript error. What do you mean by "disabling eslint anyway"? It looks like the CI runs `tsc`, which is failing (and is a helpful guard to have always passing?)

nathanstilwell: Sorry, I was thinking about this wrong earlier (I mixed up `eslint-ignore-next-line` with `@ts-ignore`). Ultimately while it may not be better than `as any` (and disabling the linter), you could just cast the `ellipsis` block as a `string`.

```jsx
ellipsis={
  (
    <>
      {"… "}
      <a
        className={cx("crl-anchor")}
        onClick={() => {
          setIsExpanded(true);
        }}
      >
        Show More
      </a>
    </>
  ) as unknown as string
}

```

PR #462, opened 2022-07-13
ui-components: add recharts and bar chart component

lassenordahl:
Currently `uPlot` is the graphing library of choice for timeseries data
across Cockroach Labs. This libraries is heavilily optimized to high
volume timeseries data, which it does great at! `uPlot` can effectively
render tens of thousands of points in a very quick timeframe.

Unfortunately! `uPlot`'s support for plugins leaves a fair bit to be
desired in terms of flexibility in visualizing categorical data. To aid
this, this PR proposes using `Recharts` with a wrapper for defining a
similarly styled graph. The first component introduced is a normal
`BarChart` component, which provides a few options for fitting an
internal use case that we have. Additional features can easily be added
through updating this component in the `ui` library.

This components functionality and a small use case is documented in the
`Charts/README.md` file, so take a look there to see how it works! This
will be an iterative process as we start to nail down what we'd like to
use.

**Quick example:**
```
 <BarChart
   width={400}
   height={400}
   margin={{ top: 32, left: 32, right: 32, bottom: 16 }}
   alignment="vertical"
   data={sampleData}
   mainKey="name"
   dataKeys={["count"]}
   barColors={["#ff0000", "#0000ff"]}
/>
```

**Quick visual example with some sample data:**
Screen Shot 2022-07-13 at 11 27 13 AM

Release note: Added Rechart support for the `ui-components` package.

# PACKAGE // TITLE

#### Checklist

### DIFF
--- a/packages/ui-components/src/Charts/barChart.module.scss
+++ b/packages/ui-components/src/Charts/barChart.module.scss
@@ -1,0 +1,28 @@
+@import "../styles/tokens.scss";
+
+.bar-chart {
+  :global(.recharts-cartesian-axis-tick-value) {
+    fill: black;
+    padding-right: 4px;
+  }
+
+  :global(.recharts-cartesian-axis-line) {
+    stroke: $color-core-neutral-3;
+    stroke-width: 2px;
+  }
+
+  :global(.recharts-cartesian-axis-tick-line) {
+    stroke: $color-core-neutral-3;
+    stroke-width: 2px;
+  }
+
+  :global(.recharts-tooltip-wrapper) {
+    border-radius: 6px;
+    overflow: hidden;
+    border: 0;
+    outline: 0;
+    background-color: $color-core-neutral-0;
+    @include crl-depth(1);
+  }
+}
+
--- a/packages/ui-components/src/Charts/barChart.tsx
+++ b/packages/ui-components/src/Charts/barChart.tsx
@@ -1,0 +1,130 @@
+import React from "react";
+import classnames from "classnames/bind";
+import {
+  BarChart as ReBarChart,
+  Bar,
+  XAxis,
+  YAxis,
+  Tooltip,
+  Legend,
+  ResponsiveContainer,
+  Cell,
+} from "recharts";
+import { Margin } from "recharts/types/util/types";
+
+import { ColorCoreNeutral0 } from "@cockroachlabs/design-tokens";
+import { METRIC_COLORS } from "./helpers";
+
+import styles from "./barChart.module.scss";
+const cx = classnames.bind(styles);
+
+export type Alignment = "vertical" | "horizontal";
+
+/**
+ * getColor is used to grab a color from a defined list of metric colors,
+ * so we can have a "semi-randomized" view of all our color schemes.
+ * @param index - Index of a metric color we want to use.
+ * @param customColors - Optional list of colors to use if available.
+ * @returns - Hex value representing a metric color.
+ */
+const getColor = (index: number, customColors?: string[]) => {
+  const colors = customColors || METRIC_COLORS;
+
+  // Default to the first color if we have an invalid index;
+  if (index < 0 || index >= colors.length) {
+    return colors[index % colors.length];
+  }
+  return colors[index];
+};
+
+interface BarChartProps<T> {
+  data: T[];
+  mainKey: keyof T;
+  dataKeys: (keyof T)[];
+  alignment: Alignment;
+  width?: number;
+  height?: number;
+  margin: Margin;
+  legend?: boolean;
+  barColors?: string[];
+  dataColors?: string[];
+}
+
+/**
+ * BarChart is a generalized wrapper with custom styling used for charting categorical
+ * data in a similar design system to the Timeseries components.
+ * @param data - Templated data object for representing the columns in the bar chart.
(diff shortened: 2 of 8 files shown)


packages/ui-components/src/Charts/barChart.module.scss
```diff
@@ -0,0 +1,38 @@
...
+    overflow: hidden;
+    border: 0;
+    outline: 0;
+    background-color: $color-core-neutral-0;
+    @include crl-depth(1);
+    @include crl-type-body;
+  }
+
+  :global(.recharts-default-tooltip) {
+    border: 0 !important;
```
nathanstilwell, line 35: Is the `!important` necessary here? If so, is there another way around it?

packages/ui-components/src/Charts/barChart.module.scss
```diff
@@ -0,0 +1,38 @@
+@import "Style/util/color.scss";
+@import "Style/util/spacing.scss";
+@import "Style/util/styling.scss";
+@import "Style/util/fonts.scss";
+@import "Style/util/type.scss";
+@import "Style/util/depth.scss";
```
nathanstilwell, line 6: So, none of this is going to work in this repo. I don't believe you are using all of these imports anyway (is that common in managed-service to over import the utilities?), but for the mixins you are using, you could port them over or use the existing design tokens and put the styles in place.

packages/ui-components/src/Charts/barChart.module.scss
```diff
@@ -0,0 +1,38 @@
...
+    stroke-width: 2px;
+  }
+
+  :global(.recharts-tooltip-wrapper) {
+    border-radius: 6px;
+    overflow: hidden;
+    border: 0;
+    outline: 0;
+    background-color: $color-core-neutral-0;
+    @include crl-depth(1);
```
nathanstilwell, line 30: `crl-depth` exists in `../styles/tokens.scss`

packages/ui-components/src/Charts/barChart.tsx
```diff
@@ -0,0 +1,127 @@
...
+          )}
+          <Tooltip
+            cursor={{ fill: ColorCoreNeutral0 }} // Prevents background highlighting of a chart.
+            itemStyle={{ display: "none" }} // Prevents a list of keys from showing.
+          />
+          {legend && <Legend />}
+          {dataKeys.map((dataKey, i) => (
+            <Bar
+              dataKey={dataKey as string}
+              fill={getColor(i, dataColors)}
```
nathanstilwell, line 110: What happens if the number of dataKeys is larger than the number of dataColors? It seems like all of the keys beyond the number of data colors will get the first color, if this expected?

nathanstilwell:
You may want to write a _Story_ for this component to make developing the component easier and provide some interactive documentation.

packages/ui-components/src/Charts/barChart.module.scss
lassenordahl: Ah whoops! Totally blanked while adding these. I removed unnecessary imports and was able to get away without one of the types mixins.

lassenordahl:
Screen Shot 2022-07-19 at 2 56 33 PM

Examples have been provided with the storybook. Going to address the remaining comments on this shortly after a meeting.

packages/ui-components/src/Charts/barChart.module.scss
lassenordahl: Removed this global override. This was necessary to override the `style` attributes applied to the tooltip, but I opted to use a `wrapperStyle` prop on the Tooltip instead.

packages/ui-components/src/Charts/barChart.tsx
lassenordahl: I opted to change it to `colors[index % colors.length];` so it'll start rotating the following colors 👍

packages/ui-components/src/Charts/barChart.tsx
```diff
@@ -0,0 +1,130 @@
...
+  return colors[index];
+};
+
+interface BarChartProps<T> {
+  data: T[];
+  mainKey: keyof T;
+  dataKeys: (keyof T)[];
+  alignment: Alignment;
+  width: number | undefined;
+  height: number | undefined;
```
nathanstilwell, line 46: Isn't the type of `number | undefined` the same an an optional parameter (`width?: number`)?

nathanstilwell (approved):
This looks great. I had one small comment, that feels like more of a nit. Nice work!

packages/ui-components/src/index.ts
```diff
@@ -10,3 +10,4 @@ export * from "./ExpandableText/ExpandableText";
 export * from "./Spinner/Spinner";
 export * from "./Typography/";
 export * from "./Logo/";
+export * from "./Charts/barChart";
```
nathanstilwell, line 13: You are missing a new line here, which is causing linting to fail.

packages/ui-components/src/Charts/barChart.tsx
lassenordahl: ah yep! updated.

packages/ui-components/src/index.ts
lassenordahl: ty!

lassenordahl:
TFTR @nathanstilwell !

PR #465, opened 2022-07-29
Upgrade Lerna to version 5.3.0

nathanstilwell:
This PR upgrades Lerna in the workspace to version 5.30. It also upgrade the github workflows to use Node version 16 (rather than the extinct Node 12) since checks were failing due to an incompatibility with Node 12 and Lerna 5+.

### DIFF
--- a/.github/workflows/deploy-storybook.yml
+++ b/.github/workflows/deploy-storybook.yml
@@ -22,7 +22,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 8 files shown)


sjbarag (approved):
Let's give it a go!

nathanstilwell:
I need to make a change to a component and would rather have these changes. `lerna run build` works locally so I'm going to merge this and try to figure out what Github's problem is later.

PR #492, opened 2022-12-02
ui-components: upgrade react/react-dom to v18

lassenordahl:
This PR just upgrades the React version used in ui-components to react 18, 
as well as the testing library version. Shouldn't have an impact on DB Console,
 since DB-Console is already using a React version newer than `ui-components`. No 
components in this Repo use new API's introduced in React 18.

Release note: None

### DIFF
--- a/packages/ui-components/package.json
+++ b/packages/ui-components/package.json
@@ -29,7 +29,10 @@
   "dependencies": {
     "@cockroachlabs/icons": "^0.5.4",
     "@popperjs/core": "^2.9.2",
+    "@testing-library/react": "^13.4.0",
     "@types/recharts": "^1.8.23",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
     "react-lines-ellipsis": "^0.15.0",
     "react-popper": "^2.2.5",
     "recharts": "^2.1.12"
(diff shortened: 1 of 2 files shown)


packages/ui-components/package.json
```diff
@@ -67,8 +70,8 @@
     "jest-environment-jsdom": "29.2.0",
     "jest-runner-eslint": "1.1.0",
     "prettier": "2.7.1",
-    "react": "17.0.2",
-    "react-dom": "17.0.2",
+    "react": "^18.2.0",
```
laurenbarker, line 73: @lassenordahl it looks like there are some dependencies listed twice.

laurenbarker (approved):
Once the dep list is cleaned up, this LGTM!

packages/ui-components/package.json
lassenordahl: I'll remove it from `devDependencies`

-- https://stackoverflow.com/questions/18875674/whats-the-difference-between-dependencies-devdependencies-and-peerdependencies

lassenordahl:
TFTR!

packages/ui-components/package.json
```diff
@@ -29,7 +29,10 @@
   "dependencies": {
     "@cockroachlabs/icons": "^0.5.4",
     "@popperjs/core": "^2.9.2",
+    "@testing-library/react": "^13.4.0",
```
laurenbarker, line 32: This should only be in dev deps since it's used for testing.

PR #496, opened 2023-01-06
[CC-6325]ui-components: remove timeout from tooltip

maryliag:
The timeout from the Tooltip component was causing the component to stick for longer than it should.
This commit removes the timeout, until a better solution is made.

Part Of https://cockroachlabs.atlassian.net/browse/CC-6325

Before
https://www.loom.com/share/3f645bba0419429f9b92d946d0538e54

After
https://www.loom.com/share/3ec548eb22624cd8b0fe543cc7d3d7b1


gtr (approved):
LGTM!

j82w (approved):

sjbarag:
CC @laurenbarker since you were just looking at tooltips

PR #506, opened 2023-03-03
Remove `-v` from release command

nathanstilwell:
Releasing seems pretty stable, so we can reduce the noise in the logs


Marjan154 (approved):
lgtm

PR #522, opened 2023-06-14
nathanstilwell/spinner updates

nathanstilwell:
- ui-components, storybook: Updating Spinner props size, intent
- storybook: playing with the storybook 7 format

### DIFF
--- a/packages/storybook-ui-components/stories/Spinner.stories.tsx
+++ b/packages/storybook-ui-components/stories/Spinner.stories.tsx
@@ -1,13 +1,30 @@
 import React from "react";
 import { withKnobs, select } from "@storybook/addon-knobs";
-import { Spinner } from "@cockroachlabs/ui-components";
+import { Spinner, SpinnerSize, SpinnerFill } from "@cockroachlabs/ui-components";
 import { StoryContainer, StoryDescription  } from "./layout";
 
+import { IconDisplaySection, IconDisplay, IconLabel, IconFrame } from "./Icon.stories";
 export default {
   title: "Spinner",
   component: Spinner,
   decorators: [withKnobs],
 };
+const sizes: Array<{ key: SpinnerSize; size: number }> = [
+  { key: "x-small", size: 16 },
+  { key: "small", size: 24 },
+  { key: "default", size: 40 },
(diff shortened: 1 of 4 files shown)


packages/storybook-ui-components/stories/Spinner.stories.tsx
```diff
@@ -1,36 +1,117 @@
...
-export default {
+import { IconDisplaySection, IconDisplay, IconLabel, IconFrame } from "./Icon.stories";
+
+const sizes: Array<{ key: SpinnerSize; size: number }> = [
+  { key: "x-small", size: 16 },
+  { key: "small", size: 24 },
+  { key: "default", size: 40 },
+  { key: "large", size: 64 },
+];
+const fills: Array<SpinnerFill> = [
```
sjbarag, line 14: nit: I think we've used `SpinnerFill[]` more consistently in the past? 🤷

packages/storybook-ui-components/stories/Spinner.stories.tsx
```diff
@@ -1,36 +1,117 @@
...
+export default meta;
+
+type Story = StoryObj<typeof Spinner>;
+
+export const Demo: Story = {
+  args: {
+    size: "default",
+    fill: "default",
+  },
+  tags: ["autodocs"],
```
sjbarag, line 56: IIRC you don't need to set `autodocs` on a particular story — it instead promotes the first story to be the "docs" story? I think? 🤔

sjbarag (commented):
Looks good!

packages/storybook-ui-components/stories/Spinner.stories.tsx
nathanstilwell: I was noticing a little weirdness in what was getting included in the "docs" story, so this might be why. Thanks for the hint.

nathanstilwell: I think you are right about only including `tags: ["autodocs"]` in the _meta_ story. At some point I'll have to come back and think about the best way to organize these stories to get the cleanest view of the autodocs feature which seems nice. I'll do some reading up on what we get with v7.

PR #531, opened 2023-07-14
Removing PR template

nathanstilwell:
This PR template was never that useful and has only been an annoyance
in recent years. As there may be significant development in this repo in
the future, I think it's time for the axe.


sjbarag (approved):
Makes perfect sense to me :)

👋 🪓

PR #536, opened 2023-09-12
deps: upgrade to typescript 5.x

sjbarag:


sjbarag:
Ah, looks like tests failed because of an eslint forwards-compat issue. We'll have to upgrade eslint to go with this.

laurenbarker (commented):
Sad 😢 I think upgrading eslint would be fine though.

sjbarag:
Approved via #538 so merging this first!

PR #537, opened 2023-09-20
deps: reduce lodash usage

sjbarag:
All of lodash.js is currently included when ui-components gets bundled, but only _.camelCase is actually necessary. Remove two unnecessary lodash functions:

* _.get was only used to provide a fallback for a top-level property, and is equivalent to Foo[bar].
* _.isEmpty was only called for a string | React.Component, and is equivalent to str !== ""

All that remains is _.camelCase, which is now imported via 'lodash/camelCase' to avoid depending on the entire lodash.js file.

### DIFF
--- a/packages/ui-components/src/Icon/CreditCard.tsx
+++ b/packages/ui-components/src/Icon/CreditCard.tsx
@@ -1,6 +1,5 @@
 import React, { useMemo, SVGProps } from "react";
 import classnames from "classnames/bind";
-import get from "lodash/get";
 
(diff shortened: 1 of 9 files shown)


laurenbarker (approved):
Yay! The less lodash the better 🎉

nathanstilwell:
Nice. If all the only dependency is `_.camelCase`, depending on the size we might just swap it out for [camelCase](https://www.npmjs.com/package/camelcase) (by sindresorhus) which is about 11k unpacked. Then we could drop lodash entirely.

PR #538, opened 2023-09-20
packaging: distribute icons and ui-components as ESM + CJS

sjbarag:
Stacks on #536, so ignore the first commit.

-----

Parcel[^1] is an alternative to webpack with a few significant differences:

1. Configuration is managed (mostly) through package.json, with minimal configuration actually needed.
2. It's significantly faster than webpack for a package of this size, as it uses native-code packages (e.g. swc) for transformation and minification
3. It has first-class support for ES Module outputs (webpack's support is still experimental)
4. It supports bundling for JS libraries in a mode that bundles only first-party code and *nothing* from node_modules.

Use parcel to dual-package ui-components as ESM + CJS, so that bundlers have access to good export analysis (via ESM) and node has access to a format that doesn't require transformation (CJS).

As a size comparison, using `du --bytes`:

```
Before (dist/main.js):           774214 B (774KB)
After (dist/cjs/index.js):        90701 B ( 90KB)
After (dist/esm/index.js):        90232 B ( 90KB)
```

This is possible because no dependencies are included in that bundle. Compared to the previous bundle, the parcel-produced bundles include zero copies of:

* lodash (or any part of it)
* recharts
* @cockroachlabs/icons
* anything else in node_modules/

Those dependencies required at runtime are all specified in package.json's 'dependencies' or 'peerDependencies' blocks. Anything in dependencies is automatically downloaded by consumers, but can be successfully de-duped with other compatible node_modules that it uses (i.e. there's no copy of lodash embedded here and another copy in the consumer's node_modules). Similarly, anything in peerDependencies must be provided explicitly by the consuming package (i.e. we declare compatibility with react 17, but installing this package won't even download one into node_modules).

The result is a JS bundle that contains only our code, minified for production use in other packages.

-----

While we're in the neighborhood and to better support dual-packaging, also upgrade recharts and set TypeScript up to use its es6 imports properly.

[^1]: https://parceljs.org


sjbarag:
Oh right I forgot to remove webpack — I'll do that in a minute.

packages/ui-components/package.json
```diff
@@ -6,11 +6,14 @@
     "type": "git",
     "url": "https://github.com/cockroachdb/ui"
   },
-  "main": "dist/main.js",
+  "source": "src/index.ts",
+  "main": "dist/cjs/index.js",
+  "module": "dist/esm/index.js",
```
sjbarag, line 11: so uh this was all the config necessary for parcel, besides the `.sassrc.json` file. It's weird, I know.

nathanstilwell:
This looks like a fun one.

sjbarag:
> This looks like a fun one.

Don't worry it's _way_ less than 3,000 lines changed. Silly `pnpm-lock.yaml` diffs.

packages/ui-components/package.json
```diff
@@ -6,11 +6,14 @@
...
   "types": "dist/types/index.d.ts",
   "scripts": {
-    "build": "npm-run-all build:typescript build:bundle",
+    "build": "npm-run-all build:typescript build:parcel",
     "build:bundle": "webpack --mode production",
```
nathanstilwell, line 15: Do we need this anymore?

packages/ui-components/package.json
laurenbarker: What is the `sassrc` file for? Is it just telling parcel that we are using sass?

nathanstilwell (approved):
Assuming this works, LGTM!!

package.json
```diff
@@ -2,8 +2,12 @@
   "name": "root",
   "private": true,
   "devDependencies": {
+    "@parcel/packager-ts": "2.9.3",
```
laurenbarker, line 5: Do parcel packages need to be installed at the root and at the package level? My knowledge of how lerna repos are organized is fairly limited...

laurenbarker (approved):
Thanks for the detailed commit message/PR description (as always). I do want to see what it looks like without webpack, but no blockers so far :lgtm:

packages/ui-components/package.json
sjbarag: Parcel automatically detects sass based on our `.ts` files importing a `.scss` file! The `.sassrc.json` fiel is mostly to make it explicit that `sass` should import packages from `node_modules`. I'll try to remove it — may not be necessary anymore.

packages/ui-components/package.json
sjbarag: Nope! Gotta remove all the webpack stuff before merging!

package.json
sjbarag: They shouldn't be necessary, but I've had a bear of a time keeping it from happen. One of parcel's selling points is that it automatically installs the plugins it needs to load the code it finds (based on the entrypoint we specify). But for some reason it installs packages in the root, not in the leaf. I'm not entirely sure why TBH, but I'll see if I can get rid of these here. AT the very least, we don't need these in both `./package.json` and `./packages/ui-components/package.json`.

packages/ui-components/package.json
sjbarag: Gone!

package.json
sjbarag: Goodness, so I managed to fix this but it required me to add a text file at `packages/ui-components/.hg` (yes, seriously) and I'm not sure if that's worse 😓

packages/ui-components/package.json
laurenbarker: Huh, that's interesting! Thanks for the explanation.

package.json
laurenbarker: I don't think it's worse! The comments in the file are very clear 🙏

laurenbarker (approved):

PR #546, opened 2024-04-09
icons: add flag icons for six new countries

rgcase:
These countries contain cloud regions that we will soon support, so we need these flags for the web client.

Release note: none


laurenbarker (approved):
Thank you! I'll go ahead and merge/publish this for testing in managed-service repo.

PR #547, opened 2024-04-11
ui-components, icons: bump package version

laurenbarker:
Bumping the version to release new flag additions.


laurenbarker:
Looks like the PR needs to can't come from a fork right now if we want a canary version published.

